PR #1, opened 2019-03-25
Try building latest OpenBLAS release: v0.3.5

tylerjereddy:
Related to discussion in: https://github.com/numpy/numpy/issues/13173

cc @matthew-brett & @charris 

Looks like I can just use the version tag for building at the exact release point based on project history. This should bump the version built from `v0.3.5.dev` (at a specific development hash) to `v0.3.5` proper.

I note that there are no badges / links in the `openblas-libs` README for monitoring the CI builds of OpenBLAS, which might be useful to have?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ env:
     global:
         # The archive that gets built has name from ``git describe`` on this
         # commit.
-        - BUILD_COMMIT=701ea88
+        - BUILD_COMMIT=v0.3.5
         - REPO_DIR=OpenBLAS
         - PLAT=x86_64
         - WHEELHOUSE_UPLOADER_USERNAME=travis-worker
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,7 @@ os: Visual Studio 2015
 
 environment:
   global:
-    OPENBLAS_COMMIT: 701ea88
+    OPENBLAS_COMMIT: v0.3.5
     OPENBLAS_ROOT: c:\opt
     MSYS2_ROOT: c:\msys64
     WHEELHOUSE_UPLOADER_USERNAME: travis-worker


tylerjereddy:
Also, since nobody has ever opened a pull request here I'm not sure if the CI is designed to handle PR OpenBLAS builds gracefully & if the cost of building a PR and then merging / building again is undesirably high.

tylerjereddy:
One other question -- are the manual processing steps for MacOS platform still needed?

matthew-brett:
I never much enjoyed those 'build-passing' badges, partly because the builds could fail for no good reason - but if you like them, please go ahead.

matthew-brett:
No - there's no need for the manual processing steps for the Mac library now, we aren't shipping dual-arch any more.

matthew-brett:
PRs are fine, I think.   It's not crazy long, and there won't be many PRs, I don't suppose.

tylerjereddy:
Builds look ok to me, but PRs don't seem to have access to the secret API key needed to upload to the server.

I suppose that makes sense--otherwise anyone could open a PR and start dumping files on the server. So, if this looks good to you, presumably it has to be merged for the uploads to happen.

matthew-brett:
Well - I think the API key is encrypted to the repo - so I *think* that PRs from the main repo, to the the same repo but master branch, will upload, but I might be wrong.

tylerjereddy:
You can check the wheelhouse upload command logs near the end of this PR CI builds vs. your last push to master to see the difference, I think. There's a warning on the PR but not on your pushes.

matthew-brett:
Right - I would have predicted that because the PR comes from your repo.  If you'd made a branch in this MacPython repo, and then done a PR from that, I would predict an upload.  But I'm speculating.

tylerjereddy:
Wouldn't I need commit rights to push the branch to the repo to open the PR? That was the motivation for the fork in the first place.

matthew-brett:
Yes, right - and you're welcome to those if you don't have them - but it would only be necessary if you needed the upload as part of the PR.  I'm guessing that usually the PRs will be getting the build steps right, and the upload will only be needed after the merge.

tylerjereddy:
That sounds right. If I work with you / Carl eventually to make less trivial adjustments to the build then uploading by default would not be desirable on more substantive PRs I suspect.

tylerjereddy:
Are you happy to merge this or do you want more changes here?

matthew-brett:
Yes, all good ..

PR #5, opened 2019-08-12
BLD: bump OpenBLAS to 0.3.7 stable

tylerjereddy:
* the stable release of OpenBLAS `0.3.7`
is now available so produce the binary
for ecosystem wheels

The problematic SkylakeX AVX512 kernels are still disabled, but there are some performance improvements elsewhere.


matthew-brett:
Merging, thanks.

PR #7, opened 2019-11-29
MAINT: bump OpenBLAS to 0.3.8.dev

tylerjereddy:
* AVX512 DGEMM kernel was rewritten
and reactivated with reasonable match
to MKL performance

* SGEMM, CGEMM, ZGEMM were adjusted
for better match to MKL performance

* there were some upstream questions
about develop branch working with NumPy,
so maybe doesn't hurt to make the binary
available in our ecosystem now; bigger
changes may be looming with new LAPACK
release recently, but OpenBLAS needs its
shims adjusted for that to be leveraged in

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ env:
     global:
         # The archive that gets built has name from ``git describe`` on this
         # commit.
-        - BUILD_COMMIT=v0.3.7
+        - BUILD_COMMIT=c815b8f
         - REPO_DIR=OpenBLAS
         - PLAT=x86_64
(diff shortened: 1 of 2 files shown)


tylerjereddy:
cc @charris @rgommers

charris:
I'd be happy to try that out, but not for the first 1.8 release.

tylerjereddy:
CI is green so I'll merge it in. This shouldn't really affect anyone unless they opt in to trying the newly-built binary of OpenBLAS.

I think bigger changes are eventually coming to the wheels infrastructure anyway, but I assume this remains relevant for now.

rgommers:
Thanks for doing this @tylerjereddy

PR #8, opened 2019-12-07
ENH: build also openblas64_ 64-bit symbol-suffixed library

pv:
Build OpenBLAS with 64-bit integer size and symbol suffix '64_' (INTERFACE64=1
SYMBOLSUFFIX=64_).

This is 64-bit BLAS/LAPACK convention avoiding symbol clashes with
32-bit BLAS/LAPACK, originally introduced for Julia.  OpenBLAS built
with this setting is also provided by some Linux distributions (e.g.
Fedora's 64-bit openblas packages).

And for clarification: the objconv tool is necessary only for OSX,
where OpenBLAS Makefiles use it for symbol renaming. As noted
in the commits, the sources are from https://www.agner.org/optimize/#objconv
and I checked the diffs vs. a few other places that are distributing
that tool.

c.f. https://github.com/numpy/numpy/pull/15012


pv:
Yes, I believe this won't change the way the 32-bit OpenBLAS libs are built. The 64-bit Windows and Linux binaries produced I think are fine. The 64-bit OSX binaries I can't check locally, apart from just the symbol names, but they're here if you want to check before merging: https://github.com/pv/openblas-libs/tree/files-osx-x86_64-1

mattip:
Any reason not to merge this?

tylerjereddy:
I was going to spot-check that generated binary locally first by running full NumPy suite with it on MacOS.

mattip:
Thanks! This is really nice work all around

pv:
Would it also be possible to push the OpenBLAS v0.3.7 builds (which is the version numpy CI is currently using)? I guess this involves pushing a commit to master.

tylerjereddy:
Let me check if I have sufficient permissions to push a feature branch to the repo & then open a PR from that---I believe that would allow the upload to occur if I understood @matthew-brett correctly a few months ago.

PR #9, opened 2019-12-17
BLD: add ppc64le, aarch64, s390x builds

mattip:
Using the recent updates to multibuild, this add multilinux2014-compatible tarballs for aarch64, ppc64le, and s390x. appveyor builds are disabled, let's see if travis likes this first


tylerjereddy:
This will be pretty cool when it is ready to go! Should remove the manual building on the other platforms, etc.

mattip:
rebased, squashed, and added more `set -v` to see why the builds are failing

mattip:
Removed the WIP label. The changes can be summarized as:
- update the multibuild repo
- update the OpenBLAS repo, even though the build checks out its own repo, it is confusing to me to have that dissonance. If reviewers request, I can revert this change
- tweak some build scripts, especially the duplicate `get_distutils_platform` function xref MacPython/gfortran-install#2
- add more architectures to the travis builds

I think this is ready for review.

mattip:
It seems strange that the other architectures all build in <10 minutes, and the x86 ones take over 30 minutes. Are there some optimizations or flags missing?

matthew-brett:
Is it just the fact that there are many more kernels for Intel?   Are the other architectures also multi-kernel?  Or are they being built for only one CPU architecture?

mattip:
Ahh, right. I guess there is no log of which variant kernels are being compiled.

mattip:
I added the _64 variants on ppc64le and s390x, for some reason the aarch64 one segfaults so I removed it for now. TBD in a future PR, maybe we need to compare with the upstream CI configuration, but this can get us started with tarballs on the rackspace instance

tylerjereddy:
Keep in mind that the tarballs will be built for v0.3.8-dev instead of the v0.3.7 stable release used in i.e., NumPy CI. I don't think it is that big of a deal, but Pauli did recently ask me to separately build the _64 ones for v0.3.7 for a better testing match. 

There is a way to do that, but I'd prefer not to unless it is really necessary, since I made a bit of a mess of it yesterday with the hash change for some v0.3.7 tarballs (they can easily get regenerated with a new hash if not careful).

tylerjereddy:
The NumPy CI would just need to be adjusted to check for the correct/bumped OpenBLAS version for those new tarballs vs. stable release tarballs in some matrix entries.

tylerjereddy:
Then when 0.3.8 stable comes out everything can sync up again.

mattip:
waiting on MacPython/gfortran-install#2

mattip:
seems to work: the aarch64 run built `openblas-v0.3.5-605-gc815b8fb-manylinux2014_aarch64.tar.gz`

mattip:
@tylerjereddy does this look ok?

PR #10, opened 2019-12-17
MAINT: build v0.3.7 binaries

tylerjereddy:
Do not merge!

* rebuild openblas-libs binaries for OpenBLAS v0.3.7
to provide this version for the 64-bit integer symbol-suffixed
builds recently added (which are otherwise currently only
available with a development commit on the path to v0.3.8)

I think because this feature branch originates in the repo proper as opposed to a fork, it should upload the new binaries to the usual rackspace location if CI passes.

I could be wrong though @pv, let me know if they don't get uploaded there after CI..

If this works, let's just close down this PR after the uploads.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ env:
     global:
         # The archive that gets built has name from ``git describe`` on this
         # commit.
-        - BUILD_COMMIT=c815b8f
+        - BUILD_COMMIT=v0.3.7
         - REPO_DIR=OpenBLAS
         - PLAT=x86_64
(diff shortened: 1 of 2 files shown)


pv:
Thanks, seemed to work fine.

The remaining build jobs could be canceled now, so that the existing Windows binaries won't be replaced (shouldn't matter otherwise but I think some places check the hash on download so it would be better to not change it).

tylerjereddy:
I'll close the PR, but I don't think I have the permissions to cancel on Appveyor. Only @matthew-brett I think.

PR #13, opened 2020-01-13
MAINT: 0.3.7 builds for new archs

tylerjereddy:
* this is NOT to be merged!

* disable OpenBLAS builds that do not
need to be regenerated, to avoid
hash changes; only retain those
matrix entries that correspond to
the new archs: ARM, POWER, s390x

* revert to BUILD_COMMIT=v0.3.7
ONLY for the purposes of this
branch/PR---objective is to generate
stable v0.3.7 builds of OpenBLAS
for downstream projects on the above
archs, since only v0.3.8 dev builds
are available at the moment

For more information see: https://github.com/numpy/numpy/pull/15279

& cc @mattip


tylerjereddy:
This has served its purpose for now I think, so closing. If we need other targeted builds we can open separate PRs for them.

mattip:
Thanks @tylerjereddy

Issue #14, opened 2020-02-09
create release artifacts of this repo


mattip:
This repo is a little different that the other MacPython wheel-producing ones. While the wheels may find a home on a PyPI-like index, the tarballs here (and perhaps the mingw ones?) I think will need a different solution. One relatively simple thing would be to create releases and host them here. Thoughts?

tylerjereddy:
I don't know much about GitHub releases so not sure I can comment on the drawbacks. Do we have a free extension with rackspace for a while now?

mattip:
Someone is doing the [releases on scipy/scipy](https://github.com/scipy/scipy/releases). Those artifacts are available via a URL like https://github.com/scipy/scipy/releases/download/v1.2.3/scipy-1.2.3-cp27-cp27mu-manylinux1_x86_64.whl which is pretty easy to parse

tylerjereddy:
Oh right, that's me. I thought you meant the new Workflows stuff

ogrisel:
Here are the docs to deploy release artifacts to github:

- for travis: https://docs.travis-ci.com/user/deployment/releases/
- for appveyor: https://www.appveyor.com/docs/deployment/github/

ogrisel:
For information, I discovered that it's possible to upload arbitrary files (that are neither wheels nor conda packages) to anaconda.org.

I started changing the CI scripts to upload anaconda artifacts for openblas-libs to anaconda.org (in the shared multibuild-wheels-staging organization):

https://github.com/ogrisel/openblas-libs/tree/anaconda-upload

I am waiting for appveyor to complete but if it works I will open a PR.

@matthew-brett you might be interested in the above as well.

tylerjereddy:
Sounds interesting. I've been working on the anaconda.org transition for scipy-wheels initially: https://github.com/MacPython/scipy-wheels/pull/68

What happens if/when we hit the 3GB free limit? For the forced replacements of wheels that may be less likely, but with a progressive accumulation of artifacts maybe it could happen, not sure.

ogrisel:
For nightly builds we should not hit the 3GB limit if we delete older wheels on a regular basis (we can do a cron job somewhere to do this).

For release staging, e.g. project can create its own organization on anaconda to get its own 2GB limit. The shared staging area is just a convenience for starting but it does not really need to be shared.

Also I was I contact with people at anaconda and they will probably raise our storage quota if we ask them as long as the bandwidth stays similar to way we had on rackspace.

[...]

Issue #4, opened 2019-05-09
OpenBLAS version number in binaries


tylerjereddy:
The binaries that are generated from this repo can, in some cases, have OpenBLAS version numbers that are not a reflection of what has actually been built.

Usually in these cases the correct commit hash on the OpenBLAS development branch will be included in the file name despite the incorrect OpenBLAS version specification.

This probably shouldn't happen--especially since we now have programmatic access to a version string directly from openblas get_config(). It causes confusion for reviewers & communication with OpenBLAS team.

tylerjereddy:
This only applies to the file names themselves--we can always know for sure with the usual ctypes calls through get_config(). Still, it can cause confusion.

mattip:
This will be closed when we move entirely to the wheels

mattip:
We have moved to wheels. Closing.

PR #17, opened 2020-02-17
Anaconda upload

ogrisel:
Use the shared staging area as openblas builds are useful for several projects.

I am not sure if the encrypted value for the token in the appveyor configuration will work. Maybe it will need to be re-encrypted by someone logged in with the MacPython github account on appveyor. It's very clear to me.

Potential fix for #14.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,10 +98,4 @@ after_success:
     # Note that the anaconda-client package on PyPI is too old. Install from github
     # tag instead:
     - pip install -q git+https://github.com/Anaconda-Platform/anaconda-client@1.7.2
-    - anaconda -t $OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN upload
-            --no-progress --force -u multibuild-wheels-staging
-            -t file -p "openblas-libs"
-            -v "$(cd OpenBLAS && git describe --tags)"
-            -d "OpenBLAS for multibuild wheels"
-            -s "OpenBLAS for multibuild wheels"
-            libs/openblas*.tar.gz
+    - upload_to_anaconda
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,7 +100,8 @@ after_success:
     - pip install -q git+https://github.com/Anaconda-Platform/anaconda-client@1.7.2
     - anaconda -t $OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN upload
             --no-progress --force -u multibuild-wheels-staging
-            -t file -p "openblas" -v "$(cd OpenBLAS && git describe --tags)"
+            -t file -p "openblas-libs"
+            -v "$(cd OpenBLAS && git describe --tags)"
             -d "OpenBLAS for multibuild wheels"
             -s "OpenBLAS for multibuild wheels"
             libs/openblas*.tar.gz
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,14 @@ env:
         # commit.
         - BUILD_COMMIT=c815b8f
         - REPO_DIR=OpenBLAS
-        - WHEELHOUSE_UPLOADER_USERNAME=travis-worker
-        # Following generated with
-        # travis encrypt -r MacPython/openblas-libs WHEELHOUSE_UPLOADER_SECRET=<the api key>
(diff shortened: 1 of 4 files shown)


ogrisel:
When I run this with my account I could upload successfully everything to the staging area:

https://anaconda.org/multibuild-wheels-staging/openblas/files

ogrisel:
I just noticed that I used `-p openblas` instead of `-p openblas-libs` to the package name. Let me know if you prefer the latter.

mattip:
Please change to openblas-libs which might make it infinitesimally more likely that someone searching for the source of the library will end up here. Even better would be if we could create a subdirectory https://anaconda.org/multibuild-wheels-staging/MacPython/openblas-libs/files, and then all the recipes under MacPython could use that staging area, making it easier for ananconda.org to manage these as a group

mattip:
Wow, when I go to the download link, I get reprimanded from Firefox that there are multiple trackers on that site, what's up with that?

mattip:
> Even better would be if we could create a subdirectory

That would probably break many things, like making the pypi-compatible index

ogrisel:
> Please change to openblas-libs which might make it infinitesimally more likely that someone searching for the source of the library will end up here.

Done.

> That would probably break many things, like making the pypi-compatible index.

For the openblas-libs binary archives, the CI consumer will use wget or curl, not pypi so there is no real point in making it pypi compatible.

For wheels on the other hand, anaconda exposes the pip compatible index URL as pypi.org would:

https://pypi.anaconda.org/scipy-wheels-nightly/simple/

> Wow, when I go to the download link, I get reprimanded from Firefox that there are multiple trackers on that site, what's up with that?

Indeed but at the same time we would not point end users to go to that page. Most downloads would happen via continuous integration bots.

ogrisel:
The travis build failed because the secret upload token is not defined in pull requests. Do you want me to change the code to skip the upload step if the secret variable is missing?

The appveyor build failed because the `MINGW_URL` download failed or was invalid. This is weird, it never failed earlier when I test on my fork.

.travis.yml
```diff
@@ -90,11 +94,14 @@ script:
...
+    # Note that the anaconda-client package on PyPI is too old. Install from github
+    # tag instead:
+    - pip install -q git+https://github.com/Anaconda-Platform/anaconda-client@1.7.2
+    - anaconda -t $OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN upload
+            --no-progress --force -u multibuild-wheels-staging
+            -t file -p "openblas-libs"
+            -v "$(cd OpenBLAS && git describe --tags)"
+            -d "OpenBLAS for multibuild wheels"
+            -s "OpenBLAS for multibuild wheels"
+            libs/openblas*.tar.gz
```
tylerjereddy: Yeah, might be useful to skip if secret variable is missing; I think that happens in some way with the previous setup as well; if you want new binaries uploaded the feature branch has to be in the repo proper rather than a PR originating from someone's fork.

That's probably good since we may not necessarily always want to change the hash on every PR that gets opened.

tylerjereddy:
Note that our version numbers on binaries are still confusing: #4. So if the uploads seem to work but with strange `v0...` that looks a bit old, it is probably ok.

ogrisel:
Yes right now the version number is grabbed by `$(cd OpenBLAS && git describe --tags)`.

travis-ci/build_steps.sh
```diff
@@ -134,3 +134,17 @@ function do_build_lib {
         $BUILD_PREFIX/lib/libopenblas* \
         $BUILD_PREFIX/lib/cmake/openblas
 }
+
+function upload_to_anaconda {
+    if [ "$OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN" == "" ]; then
+        echo "OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN is not defined: skipping."
```
ogrisel: This weird. The "after success" step passes but I cannot see this output in the travis log of the PR build.

ogrisel:
The skipping message works as expected on appveyor.

travis-ci/build_steps.sh
tylerjereddy: I can see it in the raw log:

```
+eval 'upload_to_anaconda '
OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN is not defined: skipping.
++upload_to_anaconda
++'[' '' == '' ']'
++echo 'OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN is not defined: skipping.'
```

PR #22, opened 2020-02-25
fix length of hash if used

mattip:
Previously, the hash was sometimes 8, sometimes 7 bytes. If it is used, fix it to 8. Also, as long as we are at it, create the v0.3.8 artifacts


mattip:
This should also test the new appveyor timeout settings

mattip:
maybe we should set up a PRstaging area so these build artifacts are not lost.

mattip:
0.3.8 does not build on s390x, they have a typo. Back to only the hash fix.

charris:
I'm glad we are finding and dealing with all these problems early on :)

mattip:
@tylerjereddy I am going to merge this since I can't see the upload directory name until we get a merge build. It seems the appveyor timeouts are sufficient: the first one finished.

tylerjereddy:
Not sure why you changed the openblas submodule---I imagine we can find a way to remove that submodule or at least ignore it, I never touch it when updating.

PR #25, opened 2020-03-06
update multibuild and test with xenial images

mattip:
multibuild now supports testing with xenial images. We could update across the board, this is a bit more conservative to see that everything is OK

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,35 +51,40 @@ matrix:
       env:
         - PLAT=aarch64
         - MB_ML_VER=2014
-    # causes a segfault
-    # - os: linux
-    #  arch: arm64
-    #  env:
-    #    - PLAT=aarch64
-    #    - INTERFACE64=1
-    #    - MB_ML_VER=2014
+        - DOCKER_TEST_IMAGE=multibuild/xenial_{PLAT}
+    - os: linux
+      arch: arm64
+      env:
+        - PLAT=aarch64
+        - INTERFACE64=1
+        - MB_ML_VER=2014
+        - DOCKER_TEST_IMAGE=multibuild/xenial_{PLAT}
     - os: linux
       arch: s390x
       env:
         - PLAT=s390x
         - MB_ML_VER=2014
+        - DOCKER_TEST_IMAGE=multibuild/xenial_{PLAT}
     - os: linux
       arch: s390x
       env:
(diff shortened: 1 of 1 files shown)


mattip:
I am not even sure we use `install_run` on this repo ...

.travis.yml
```diff
@@ -51,35 +51,40 @@ matrix:
...
         - MB_ML_VER=2014
-    # causes a segfault
-    # - os: linux
-    #  arch: arm64
-    #  env:
-    #    - PLAT=aarch64
-    #    - INTERFACE64=1
-    #    - MB_ML_VER=2014
+        - DOCKER_TEST_IMAGE=multibuild/xenial_{PLAT}
+     - os: linux
```
tylerjereddy: There's an extra space before the `- os` that is causing the Travis CI yml parser to fail, so Travis CI never runs for this PR. Checked by pasting into their linter here: https://config.travis-ci.com/explore

mattip: fixing

mattip:
Yes, it should fix the "ContextManager" problem.

tylerjereddy:
I restarted the two failing Travis CI matrix entries. One was a "no output received timeout" & the other was the s390x "no space left" issue we've seen from time to time.

If this becomes all green after the restarts I'll try to merge promptly.

.travis.yml
```diff
@@ -51,35 +51,40 @@ matrix:
...
-    #    - INTERFACE64=1
-    #    - MB_ML_VER=2014
+        - DOCKER_TEST_IMAGE=multibuild/xenial_{PLAT}
+    - os: linux
+      arch: arm64
+      env:
+        - PLAT=aarch64
+        - INTERFACE64=1
+        - MB_ML_VER=2014
+        - DOCKER_TEST_IMAGE=multibuild/xenial_{PLAT}
```
tylerjereddy: This job is still segfaulting--should we disable for now & perhaps try again when we bump up to `0.3.9` or beyond?

mattip: yes. Should we report to upstream?

tylerjereddy: If it keeps happening after I rebase the 0.3.9 bump I'll report it yeah

tylerjereddy:
CI all green now, looks good, thanks Matti.

PR #26, opened 2020-03-08
MAINT: restore ARM64 64-bit int

tylerjereddy:
* restore `arm64` 64-bit integer OpenBLAS
build in Travis CI matrix now that we
are building OpenBLAS 0.3.9 stable

* if this Travis CI matrix entry continues
to segfault during the build of OpenBLAS
we should report it upstream

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,13 +52,13 @@ matrix:
         - PLAT=aarch64
         - MB_ML_VER=2014
         - DOCKER_TEST_IMAGE=multibuild/xenial_{PLAT}
-    #- os: linux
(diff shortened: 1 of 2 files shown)


tylerjereddy:
Reported upstream as you can see in the cross-link above

tylerjereddy:
Issue has been fixed upstream so I'll need to bump OpenBLAS hash again. Will try that soon-ish.

tylerjereddy:
Ok, we're good to go!  The full build matrix has been restored and CI is green. We do have some exotic archs like 390x that sometimes need restarts, but good for now...

PR #29, opened 2020-03-30
DRAFT: Update openblas

mattip:
try out some new code to maybe fix the dynamic function choice. I am trying to get the artifacts uploaded to test numpy builds with them


mattip:
Closing. The needed artifacts were uploaded. Only the 64-bit s390x run succeeded, so I used the 64-bit artifact in numpy/numpy#15868

isuruf:
This might fix the current manifestation of this issue, but the underlying problem is not fixed. You need to set `TARGET` explicitly.

PR #31, opened 2020-04-01
update to latest develop, add TARGETs

mattip:
add minimal build targets for the various platforms


mattip:
@isuruf what should the TARGET be on s390x, ppc64le? I used PRESCOTT for x86_32 and x86_64, and ARMV8 on aarch64.

isuruf:
For ppc64le, `POWER8`.
For s390x `ZARCH_GENERIC` or `Z13` or `Z14`. Not sure which one is appropriate.
For `x86`, I think `KATMAI`, but not sure.

tylerjereddy:
Note that artifacts from this PR may be uploaded as you iterate here since the feature branch is on the repo itself.

mattip:
closing, will re-submit via a fork

PR #34, opened 2020-07-10
add TARGETs

mattip:
OpenBLAS reccomends using a minimum build `TARGET` when compiling with `DYNAMIC_ARCH=1`. Otherwise the TARGET is the build machine, which may set the bar too high for the runtime machine. Maybe connected to some of the segfaults we see on random runs of CI?


mattip:
xref #31 for more info

charris:
Seems reasonable.

mattip:
Note that merging this will upload new tarballs, which I think will break numpy. I am not sure how to work around this: maybe we should first disable the hashes in numpy, thenmerge this, then update the hashes in numpy.

tylerjereddy:
SciPy started using that hash scheme too, although I've not been able so get `0.3.10` adopted in SciPy `master` just yet because of strange ilp64 test issues on windows.

mattip:
@charris ok t merge this?

charris:
@mattip I merged it :) The master branch here is for experimentation. Thanks Matti. Probably not worth worrying about a backport to 1.19. What do you think?

mattip:
Probably not worth the effort to backport. We haven't gotten any reports in the wild about crashes due to TARGET not being set. This whole exercise is to eliminate a niggle of concern about the strange segfaults we see sometimes in PyPy, but who knows what is really going on there.

charris:
My main concern is that we don't want to set the bar too low. OTOH, if the wheel builds could  depend on factors out of our control, like the build architecture, that would not be good either. Is it the case that OpenBLAS does runtime checks for architecture support or is it all compile time?

mattip:
From what I understand the TARGET is the minimum CPU supported, then the DYNAMIC_ARCH=1 flag triggers more optimal loops based on runtime detection.

mattip:
We can compare 1.19 benchmarks to HEAD to see if there is a degradation

PR #35, opened 2020-07-24
try commit 349b722d8d09303

mattip:
Build openblas libraries for 349b722d8d09303 to maybe fix numpy/numpy#16913

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ env:
     global:
         # The archive that gets built has name from ``git describe`` on this
         # commit.
-        - BUILD_COMMIT=v0.3.10
+        - BUILD_COMMIT=349b722d8d09303
         - REPO_DIR=OpenBLAS
         # Following generated with:
         # travis encrypt -r MacPython/openblas-libs OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN=<secret token value>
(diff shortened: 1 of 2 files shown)


bashtage:
You might need to make -j 2 on appveyor to avoid time out.  I think each AppVeyor instance has 2 real CPUs.

mattip:
Appveyor succeeded, macOS on travis failed. It seem it miscompiled somehow: did not properly generate config_kernel.h. Is there a way to restart just that travis run?

bashtage:
If you are an owner you can log into Travis and you should see a restart icon on the job.

image

bashtage:
Somehow I am an owner, so I restarted it.

mattip:
build succeeded. I will self-merge to move this forward. It should not delete the current artifacts at https://anaconda.org/multibuild-wheels-staging/openblas-libs/files, only add to them. Then the next step is to make a PR to numpy/numpy to update tools/openblas_support.py, updating OPENBLAS_V to the string inside the DLL and OPENBLAS_LONG to the name of the download file. In order to avoid the problem of updating all the hashes, the check for `if sha256_returned != sha256_expected` can be disabled. Much of this all can be tested with `python tools/openblas_support.py --test`

PR #37, opened 2020-09-17
add manylinux2014 build targets for linux

mattip:
we have manylinux1, manylinux2010, but not manylinux2014 x86 tarballs.

xref matthew-brett/multibuild#355

At some point we could rename these to not use manylinux tags to save the triple builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,6 +100,21 @@ matrix:
         - INTERFACE64=1
         - MB_ML_VER=2010
 
+    # linux manylinux2014
+    - os: linux
+      env:
+        - PLAT=x86_64
+        - MB_ML_VER=2014
+    - os: linux
+      env:
+        - PLAT=i686
+        - MB_ML_VER=2014
+    - os: linux
+      env:
+        - PLAT=x86_64
+        - INTERFACE64=1
+        - MB_ML_VER=2014
+
 before_install:
     - source travis-ci/build_steps.sh
     - before_build


mattip:
Seems appveyor is failing due to this issue msys2/MSYS2-packages#2058. Tangled web indeed. Luckily , appveyor is not needed for this PR which only involves linux

.travis.yml
```diff
@@ -100,6 +100,21 @@ matrix:
...
+        - PLAT=x86_64
+        - MB_ML_VER=2014
+    - os: linux
+      env:
+        - PLAT=i686
+        - MB_ML_VER=2014
+    - os: linux
+      env:
+        - PLAT=x86_64
+        - INTERFACE64=1
```
charris, line 115: Do we have a way to distinguish the interface64 wheels for download?

charris: NVM, I didn't notice this was for building the libraries.

tylerjereddy (commented):
Any reason to update the OpenBLAS submodule? I thought we were pulling in the version/hash based on `BUILD_COMMIT`?

mattip:
> Any reason to update the OpenBLAS submodule

No, that was a mistake, thanks.

tylerjereddy:
Alright, I'll put this in.

I believe NumPy is using an even newer version of OpenBLAS in `openblas_support.py` (the longer version suggests you needed a patch after the stable release) so the hash regenerations for `v0.3.10` may not be too disruptive (although if you do a point release of NumPy with an older version of `openblas_support.py` it is possible that NumPy wheel builds might need a backport update to the hashes).

SciPy is still on OpenBLAS `v0.3.9` so should not be disruptive there.

tylerjereddy:
Thanks @mattip

radarhere:
So I presume this won't actually appear at https://anaconda.org/multibuild-wheels-staging/openblas-libs/files (without a hash) until the next version of OpenBLAS is released. Just asking, is there an estimate for when that would be?

mattip:
The newer one already appears https://anaconda.org/multibuild-wheels-staging/openblas-libs/v0.3.9-452-g349b722d/download/openblas-v0.3.9-452-g349b722d-manylinux2014_x86_64.tar.gz. For matthew-brett/multibuild#355 we should also rebuild v0.3.10, which I thought this PR would do.

mattip:
@radarhere I created the v0.3.10 manylinux2014 tarballs as well. Thanks for pointing it out.

Issue #38, opened 2020-09-20
CI: MSYS2 Appveyor CI failures


tylerjereddy:
A package manager/leadership changeover in the MSYS2 project is apparently causing CI failures for us here.

The news from their team: https://www.msys2.org/news/#2020-06-29-new-packagers 

Example failure from Matti's recent pull request #37: https://ci.appveyor.com/project/tylerjereddy/openblas-libs/builds/35273592/job/qc414wkexr0k2oq7

tylerjereddy:
Closing because we switched to GitHub actions for windows builds in #41.

PR #41, opened 2020-10-14
Use GitHub actions on windows instead of appveyor

mattip:
github actions gives us up to 6 hours, the builds take ~1 hour each ( X3: win32, win64, win64 w/64-bit interface)


mattip:
I haven't enabled and checked uploads yet

.github/workflows/build.yml
```diff
@@ -2,33 +2,83 @@
...
+          fi
+          echo "START_DIR=$PWD" >> $GITHUB_ENV;
+          choco install -y zip
 
-      # Runs a set of commands using the runners shell
-      - name: Run a multi-line script
+      - run: |
+          choco install -y mingw --forcex86 --force --version=7.3.0
+          refreshenv
+          echo "C:\\ProgramData\\chocolatey\\lib\\mingw\\tools\\install\\mingw32\\bin" >> $GITHUB_PATH;
```
mattip, line 52: This part matches the numpy/numpy build commands. The path manipulation didn't work, I should delete it.

.github/workflows/build.yml
```diff
@@ -2,33 +2,83 @@
...
+        name: Install 32-bit mingw
+        shell: powershell
+        if: ${{ matrix.BUILD_BITS == '32' }}
+
+      - name: Build
+        run: |
+          BITS=${{ matrix.BUILD_BITS }}
+          if [ "$BITS" == "32" ]; then
+            export PATH=/c/ProgramData/chocolatey/lib/mingw/tools/install/mingw32/bin:$PATH
+          fi
```
mattip, line 62: This path manipulation works. The problem is that the github actions bash runner prepends `/bin/mingw64` to the path no matter what `$GITHUB_PATH` says to do.

.github/workflows/build.yml
mattip: done

mattip:
[The secret](https://github.com/MacPython/openblas-libs/pull/41/checks?check_run_id=1252508837#step:8:36) will only be available after merge, so the last part is hard to test

mattip:
@charris, @tylerjereddy: any reason not to merge?

tylerjereddy (approved):
Makes sense, can't really argue against it given that our Windows/Appveyor builds are bricked by #38 and it probably won't be long before we'll need to start producing more binaries so best to know sooner than later if there are issues with this migration.

Thanks Matti

Issue #42, opened 2020-10-16
Windows uploads are not picking up the 64-bit interface difference


mattip:
After the move to github actions, [the upload](https://github.com/MacPython/openblas-libs/runs/1261972663?check_suite_focus=true#step:8:40) on windows64, interface=64 is using 
```
openblas-v0.3.9-452-g349b722d-win_amd64-gcc_8_1_0.zip
```
and so is the [upload for](https://github.com/MacPython/openblas-libs/runs/1261972678?check_suite_focus=true#step:8:45) windows64, interface-32 (which warns the file already exists).

Note the new compiler tag. For 32-bit builds it is 
```
openblas-v0.3.9-452-g349b722d-win32-gcc_7_3_0.zip
```

so we will need to adjust NumPy/SciPy to accommodate that change

PR #43, opened 2020-10-16
Update version to v0.3.11, Fix INTERFACE64, gcc version

mattip:
Fixes gh-42. There must be a better way to get the matrix values into the environment, but this should work.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ on:
     branches: [ master ]
 
 env:
-  OPENBLAS_COMMIT: 349b722d8d09303
+  OPENBLAS_COMMIT: "v0.3.11"
   OPENBLAS_ROOT: "c:\\opt"
 
 jobs:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ on:
     branches: [ master ]
 
 env:
-  OPENBLAS_COMMIT: "v0.3.11"
+  OPENBLAS_COMMIT: "14b1d339331d"
   OPENBLAS_ROOT: "c:\\opt"
 
 jobs:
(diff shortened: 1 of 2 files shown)


mattip:
Merging this will change the hashes for the already-uploaded packages. Should I go ahead and get the HEAD from OpenBLAS so we can check their fix for numpy/numpy#16774 ?

mattip:
... and of course I have a typo.

mattip:
gcc fails to compile 0.3.11 since the mingw version is 8.1.0 which suffers from [this bug](https://gcc.gnu.org/git/?p=gcc.git&a=commit;h=dd341e6b9a8254a6b41b146afce0917889f5bb6a): 
```
config/i386/avx512fintrin.h (_mm512_abs_pd, _mm512_mask_abs_pd): Use
__m512d type for __A argument rather than __m512
```

It doesn't appear that there are later versions of mingw available.

mattip:
[skip travis]

mattip:
Now getting
```
undefined reference to `dgetri_`
```
in 32-bit tests. When do we drop windows 32-bits?

tylerjereddy:
> When do we drop windows 32-bits?

My impression was that Python 32-bit interpreter installs on Windows were still very common for whatever reason. I think Matthew Brett may have been saying that?

charris:
> I think Matthew Brett may have been saying that?

That's also my recollection.

mattip:
I was blaming the 32-bit build too soon. I set `fail-fast` to false so all builds run to completion, and they all give the `undefined reference to dgetri_` error. I am trying to reproduce locally.

matthew-brett:
About importance of Windows 32-bit bit - yes, that is usually me quoting Steve Dower : https://github.com/pypa/manylinux/issues/118#issuecomment-481404761 . I haven't investigated the wheel download statistics though.

mattip:
something is still off with the interface64 builds, the internal OpenBLAS linktest is failing.

mattip:
self-merging to move this chain of fixes forward

PR #44, opened 2020-10-22
specify python version for macos

mattip:
specify python version on macosx


mattip:
I don't know if specifying the python version in the build step persists into the upload, but we will find out when I merge this.

PR #46, opened 2020-11-12
WIP: Build with BUFFERSIZE=20

mattip:
Add BUFFER_SIZE=20 to the build. Mark the tarballs with this option so we can try it out.


mattip:
Marked as a WIP, the artifacts will be uploaded since I made the branch in this repo and we can use them in testing once they are up. If we like it, I will remove the `buffersize=20` suffix before we merge this

mattip:
Closing. Done differently in gh-48

PR #47, opened 2020-12-15
MAINT: build v0.3.13

tylerjereddy:
* build latest version of OpenBLAS
released two days ago

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ on:
     branches: [ master ]
 
 env:
-  OPENBLAS_COMMIT: "v0.3.12"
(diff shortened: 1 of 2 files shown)


mattip:
MacOS has a problem: `getarch_2nd.c:5:10: fatal error: 'config_kernel.h' file not found`. We should migrate all the macOS and linux jobs we can off travis as well.

mattip:
This should also get the changes from the BUFFERSIZE=20 PR but without changing the upload name

mattip:
Closing, done with a move to github actions for x86 and x86_64 machines in gh-48. Also used BUFFERSIZE=20 for all builds.

PR #48, opened 2020-12-15
move builds to github actions, update to BUFFERSIZE=20 and OpenBLAS v0.3.13

mattip:
Things to undo once this works
- move travis to legacy until this is merged
- the upload is without --force to avoid too much churn


mattip:
There are 15 builds in the new task  (3 python versions) * (`macOS` + `macOS,interface64` + `linux64,interface64` + `linux64` + `linux32`)

mattip:
The upload step succeeds, using a `--skip` directive to not override existing tarballs. 

The calculation in the comment above was wrong: there are 5 tarballs needed, the python version does not matter for OpenBLAS. I was getting this PR confused with other ones for MacPython that produce wheels.

But there is are two problems: the interface64 is not appearing in the label and I am not producing manylinux2010 tarballs.

mattip:
That is better: now there are (`manylinux1`, `manylinux2010`, `manylinux2014`) * (`ubuntu_x86`, `ubuntu_x86_64`, `ubuntu_x86_64 + interface64`) + `macOS` + `macOS +  interface64` = 11 jobs

mattip:
The 0.3.13 tarballs were uploaded for the various x86 platforms. Now to re-enable a more limited travis build.

PR #54, opened 2021-02-28
Manylinux 2 24

mattip:
Add a manylinux_2_24 build and update multibuild

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,4 @@
-# This is a basic workflow to help you get started with Actions
-
-name: Windows build
+name: Windows
 
 defaults:
   run:
--- a/.github/workflows/multibuild.yml
+++ b/.github/workflows/multibuild.yml
@@ -19,7 +19,7 @@ jobs:
         platform: [x64]
         PLAT: [i686, x86_64]
         INTERFACE64: ['', '1']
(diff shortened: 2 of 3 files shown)


tylerjereddy (commented):
Not sure if it helps, but looks like CI may be choking because of this and similar `multibuild` lines: https://github.com/matthew-brett/multibuild/blob/devel/common_utils.sh#L343

I checked the same docker image locally and `python3` works with that command, but not `python` itself.

mattip:
> I checked the same docker image locally and python3 works with that command, but not python itself.

Thanks. The image is based on Debian not CentOS, and they do not install python by default. I will make a PR to multibuild.

mattip:
It seems we need support for gcc7+ to build OpenBLAS, manylinux_2_24 comes with gcc 6.3.10. I will try to update the manylinux image.

mattip:
ahh, [not so simple](https://github.com/pypa/manylinux/issues/1012#issuecomment-787030543). Updating gcc on debian will also update libgcc++. :(

matthew-brett:
Is it practical to compile gcc?

mattip:
I guess that is what needs to be done, in a way that preserves the system libraries so they will be "portable" to standard debian9 machines.

mattip:
Closing as abandoned.

PR #56, opened 2021-03-03
fix github actions uplaod

isuruf:
cc @mattip


mattip:
Since master is broken might as well merge this without waiting. Thanks @isuruf

mattip:
... and the branch build would not hit the upload code anyway

PR #57, opened 2021-04-10
manylinux_2_24 gcc-9

mayeut:
c.f. #54 & https://github.com/pypa/manylinux/issues/1012#issuecomment-817025210

Install gcc-9 from ubuntu-toolchain-r/test PPA to build OpenBLAS


mattip:
There is still a problem on ppc64, [`dynamic_power.c:206:6: error: cpu ‘power10’ is an invalid argument to builtin ‘__builtin_cpu_is’`](https://travis-ci.com/github/MacPython/openblas-libs/jobs/499253485#L605). See PR xianyi/OpenBLAS#2994 which caused the problem, and issue xianyi/OpenBLAS#3074 which found it. This was fixed in `b94dab525` for OpenBLAS. We have OpenBLAS pinned to an earlier commit `af2b0d0`. So maybe try modifying the `.travis.yml` and `.github/workflows/*.yml` files to use the later changeset.

mattip:
All checks passed :tada:. Do you want to remove the "Draft" status?

mayeut:
@mattip,

Sorry but I kind of hi-jacked this PR to see what could be done with custom PPA...
The answer is PPA are probably the way to go to get newer tools in manylinux_2_24 image. However, the quality of the "http://ppa.launchpad.net/mayeut-github/test*" PPA used in this PR is not good at all (I merely tried stuff to get builds passing without much knowledge regarding what I was actually changing in the package source...). I'll try to sum-up what I tried in https://github.com/pypa/manylinux/issues/1012 and hopefully, someone with knowledge of debian packaging can take over providing a PPA with at least binutils. If the gcc-10 can also be patched with compatibility for stdc++ like devtoolset is on RH, then it would certainly be interesting to include this in the stock manylinux images.

I do feel comfortable using the gcc-9 PPA from the debian/ubuntu folks and rebuilding binutils from source as was once tried in this PR. As you noted, it will require bumping OpenBLAS to get past the power10 issue on ppc64le. I will update the PR this way.

mattip:
> I will update the PR this way.

Thanks!

mayeut:
I cleaned this up a bit.
There are 2 errored jobs on ppc64le. I'm confident they'll pass given the outcome of the other jobs.
Seems GitHub changed the way Actions are triggered and thus those workflows weren't triggered with the last update.

tylerjereddy:
> Seems GitHub changed the way Actions are triggered

I think it is a protection against cryptocurrency mining (seriously). Anyway, I approved the CI to run--I think this only happens first time for new contributor.

mattip:
I think this should be good to go. I am not sure what to do about the g++ compatibility issue, it will be a blocker for actually releasing manylinux2_24 wheels, since SciPy, and soon NumPy, use c++ code.

mattip:
Closing. the g++ issue prevents manylinux2_24 wheels from being viable.

PR #69, opened 2021-12-25
Use OpenBLAS v0.3.19

DWesl:


tylerjereddy:
Ah, unfortunately some CI failures. Perhaps not surprising given all of the ecosystem gymnastics that have had to happen to deal with the M1 Mac/arm64 stuff.

DWesl:
It looks like the MacOS arm64 runs were trying to use an XCode version that had been dropped from the runner.  The MacOS x64/x86 builder is complaining it can't find `-lSystem`, which I don't know how to fix.

mattip:
[This answer](https://developer.apple.com/forums/thread/666700) suggests setting LDFLAGS
```
export LDFLAGS=-L/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib
```

mattip:
Thanks @DWesl. The files should appear on https://anaconda.org/multibuild-wheels-staging/openblas-libs/files

PR #71, opened 2022-02-13
Add a manylinux_2_28 build

mayeut:
Per https://github.com/pypa/manylinux/pull/1277#issuecomment-1038218157 comment 
Draft to check if manylinux_2_28 Proof-Of-Concept images can build openblas-libs.
This reduce testing to manylinux_2_28 at first. If all is OK, will preextend the build to check nothing broke for other builds.

cc @mattip

### DIFF
--- a/.github/workflows/multibuild.yml
+++ b/.github/workflows/multibuild.yml
@@ -17,11 +17,9 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
         platform: [x64]
-        #PLAT: [i686, x86_64]
-        PLAT: [x86_64]
+        PLAT: [i686, x86_64]
         INTERFACE64: ['', '1']
-        #MB_ML_VER: ['', 2010, 2014, '_2_28']
-        MB_ML_VER: ['_2_28']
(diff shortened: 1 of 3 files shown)


mattip:
Looks good.

mayeut:
Re-enable all builds then.

@mattip, do you know if there's an easy way to enable some testing ? Workflows seem to only do the build, no tests. Am I missing something ?

mattip:
We should move the linux and macOS runs off travis and onto github actions, and enable the tests.

mayeut:
> We should move the linux and macOS runs off travis and onto github actions

This is already the case except for alt-arch which are running on travis-ci and that probably should remain like this (going with qemu on GHA might really prove too slow).

> enable the tests

any insight on this one ?

mattip:
manylinux_2_17 will EOL in June 2024. We should restart this work to support manylinux_2_28 across the scientific python ecosystem. I think this as good a place as any to start

charris:
> We should restart this work to support manylinux_2_28

I was just thinking the same. We've been using manylinux2014 for a _long_ time. It looks like 2_28 doesn't support 32 bit linux, so the two decisions are coupled.

PR #73, opened 2022-03-16
Musllinux

### LINKED ISSUE #72: Build for musl linux
In order to build musl linux artifacts, we need to add it to the [github action](https://github.com/MacPython/openblas-libs/blob/master/.github/workflows/multibuild.yml) which uses multibuild. But the Pr to add musl to multibuild seems stalled multi-build/multibuild#430

mattip:
Closes #72 

Also updates the multibuild subrepo to point to https://github.com/multi-build/multibuild, which means you need to run `git submodule sync` in your fork

### DIFF
--- a/travis-ci/build_steps.sh
+++ b/travis-ci/build_steps.sh
@@ -85,6 +85,7 @@ function build_lib {
         -e INTERFACE64="${interface64}" \
         -e PYTHON_VERSION="$MB_PYTHON_VERSION" \
         -e MB_ML_VER=${manylinux} \
+        -e MB_ML_LIBC=${libc} \
         -v $PWD:/io \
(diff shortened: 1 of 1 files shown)


mattip:
The build is trying to use a `manylinux_1_1_x86_64` docker image, which is wrong.

mattip:
@liath do you see where the problem is with the [musllinux stanza](https://github.com/MacPython/openblas-libs/pull/73/files#diff-987f3c7ab4eb07824fa0529bff2336e90868217aa1699d7757bf5f2b0ecbffa9R32)

liath:
Ah sorry, just noticed this. Looks like you got it sorted! Thanks :)

Issue #68, opened 2021-12-06
Consider modern mingw-w64 UCRT build


matthew-brett:
Building Scipy with Meson, I found I needed to rebuild OpenBLAS with mingw-w64, using the new UCRT linkage.

https://github.com/matthew-brett/openblas-libs/blob/ucrt-build/build_openblas.ps1

In particular, I got the compiler from the [rtools](https://github.com/r-windows/rtools-ucrt), installed with:

```
choco install rtools
```

but that is, I believe, equivalent to the latest mingw-w64 UCRT compiler chain.

Would it be reasonable to use that toolchain for the standard CI compile here?

mattip:
We have moved our windows builds to visual 2017 and are on the cusp of moving to 2019, so as I understand UCRT would be fine.

matthew-brett:
I think I've cracked this one, with #82

mattip:
Closing, #82 was merged

PR #75, opened 2022-05-18
Build openblas with ucrt mingw

mattip:
Creating PR so we can move this forward


mattip:
@matthew-brett I am not sure how this works: all I see is a new script. Does multibuild automatically use `build_openblas.ps1` somehow? Do we need the toolchain mentioned in #68?

mattip:
Closing in favor of #76

PR #76, opened 2022-05-18
Build openblas with ucrt mingw

### LINKED ISSUE #68: Consider modern mingw-w64 UCRT build

### LINKED ISSUE #74: pkgconfig files are missing in the built package
See https://github.com/scipy/scipy/pull/16187#issuecomment-1129166785 for more context. All development headers and CMake files are present, but `.pc` files are missing. We need those for building with Meson.

mattip:
Replaces #75 since I don't have push permissions to that repo. 

Fixes #68 fixes #74

I don't understand some of the differences between `build_openblas.ps1` and `build_openblas.sh`. Some of the variable names have changed, there is some renaming that is different, and it seems the zip step at the end is missing.


matthew-brett:
Yes - sorry - the `.ps1` file was just me working it out as I went.    It really needs a rewrite, and it will probably work as a refactor of the `.sh` script.  Yes, it does need the `-ucrt` toolchain - Scipy is using `rtools-4` from `choco` - I believe.

carlkl:
AFAIK OpenBLAS still uses long double for intermediate calculations. i.e.: interface/zrotg.c
So using the $long_double="mlong-double-64" flag on Windows is not recommended without patching OpenBLAS.
See i.e. https://github.com/xianyi/OpenBLAS/issues/1711 and https://github.com/xianyi/OpenBLAS/issues/1270. See also this comment: https://github.com/xianyi/OpenBLAS/issues/1270#issuecomment-334980009
The _non-standard toolchain called mingwpy_ was at this time a mingw-w64 toolchain patched to use "mlong-double-64" as default.

mattip:
Closing in favor of #82

PR #77, opened 2022-05-19
Pkgconfig

### LINKED ISSUE #74: pkgconfig files are missing in the built package

mattip:
Add a pkgconfig file to the tarballs. Fixes #74. Note that updating the submodule to OpenBLAS 0.3.20 does __not__ update the built OpenBLAS to 0.3.20, it just makes local developement easier.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,6 +91,10 @@ jobs:
           cp $(cygpath $OPENBLAS_ROOT)/$BITS/bin/*.dll .
           ./test_dyn.exe
 
+      - uses: actions/upload-artifact@v3
+        with:
+          path: builds/openblas*.zip
+
       - name: Upload
         env:
          OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN: ${{ secrets.MULTIBUILD_WHEELS_STAGING_ACCESS }}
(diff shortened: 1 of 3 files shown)


mattip:
The `openblas.pc` file assumes the tarball will be unpacked into `C:\opt` on windows and `/usr/local` otherwise. We should document the need to set `PKG_CONFIG_PATH` to find it. Also, local builds will prefer to set `PKG_CONFIG_SYSROOT_DIR` since they may not be able to unpack into those locations.

mattip:
@rgommers: do you want this for the current OpenBLAS release? If I merge this it will overwrite the existing files.

matthew-brett:
Do all the binaries work?   If so - then yes, let's use them.

mattip:
gahh, the uploaded file name doesn't have the version anymore it is `[openblas64_--manylinux2014_x86_64.tar.gz](https://anaconda.org/multibuild-wheels-staging/openblas-libs/v0.3.19-22-g5188aede/download/openblas64_--manylinux2014_x86_64.tar.gz)

mattip:
[`fatal: unsafe repository ('/io/OpenBLAS' is owned by someone else)`](https://github.com/MacPython/openblas-libs/pull/77/checks#step:7:48100)

PR #78, opened 2022-05-19
remove unused appveyor, update to OpenBLAS v0.3.20

mattip:
OpenBLAS v0.3.20 was released Feb 20, let's build for it.


mattip:
I checked the artifacts to make sure they have the correct version tag, and that they have the pkgconfig file.

PR #80, opened 2022-06-09
use -f"visibility-protected" when building

### LINKED ISSUE #79: Hide symbols that are not meant ot be public
Over at numpy/numpy#21643 it turns out that when we use `SYMBOLSUFFIX` not all the exported symbols are converted. For instance, in NumPy 1.22.4, we have `cblas_dgemm64_` calling `dgemm_nn` which does not have the `64_` suffix:
```
$ objdump /tmp/venv/lib/python3.10/site-packages/numpy.libs/libopenblas64_p-r0-2f7c42d4.3.18.so -T | grep dgemm_nn
00000000001fdb40 g    DF .text	0000000000000638  Base        dgemm_nn
```

This is problematic if there is a different openblas.so already loaded into the process. The first one's symbols (which could be using a 32-bit interface) will be used, which can segfault.

Should we be adding `CFLAGS=-fvisibility=hidden to [our builds](https://github.com/MacPython/openblas-libs/blob/cc761826c220cb2c44b82d8dee4c11c3db44cae6/travis-ci/build_steps.sh#L166)?

@martin-frbg thoughts?

mattip:
Fixes #79

Also update OpenBLAS so the upload will not overwrite existing files

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ on:
     branches: [ master ]
 
 env:
-  OPENBLAS_COMMIT: "v0.3.20"
+  OPENBLAS_COMMIT: "bfd9c1b58cd3"
   OPENBLAS_ROOT: "c:\\opt"
 
 jobs:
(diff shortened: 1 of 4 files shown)


mattip:
We should add an artifact upload step so we can see the results before merging

mattip:
Ahh, I did already add an artifact upload step :). The artifacts are only availalble at the end of the github action run.

mattip:
Building locally with the artifacts solved the segfault.

PR #81, opened 2022-06-10
backport PR 80 to v0.3.20

mattip:
This will overwrite the https://anaconda.org/multibuild-wheels-staging/openblas-libs/files with the version including #80 to use `protected` with the posix builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ on:
     branches: [ master ]
 
 env:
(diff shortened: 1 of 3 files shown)


mattip:
Somehow just setting CFLAGS seems to [break arm64-apple-darwin20.0.0-gfortran](https://github.com/MacPython/openblas-libs/runs/6832414065?check_suite_focus=true#step:7:1326)

PR #84, opened 2022-06-23
Fix for CFLAGS

matthew-brett:
The following line is wiping out the set `CFLAGS` for the cross-compile:

```
      && CFLAGS="-fvisibility=protected" make BUFFERSIZE=20 ...
```

https://github.com/MacPython/openblas-libs/runs/6817632611

Put the existing CFLAGS back.


matthew-brett:
Merging - as I think this one is good, and no comments after a longish while.

PR #85, opened 2022-06-25
WIP: patch OpenBLAS build for quadmath symbols

matthew-brett:
Apply Carl's patch to OpenBLAS export build of library, to remove
libquatmath symbol.

See:
https://github.com/MacPython/openblas-libs/pull/82#issuecomment-1161869998

Disabled for 32-bit (symbol missing).

We should try and apply this for Unix builds.


mattip:
@matthew-brett I am going to push forward with this to solve https://github.com/numpy/numpy/issues/27080. How can I make sure no quadmath symbols are used in the static compilation of libscipy-openblas.dll?

matthew-brett:
@carlkl  - this was your patch - any thoughts on - for example - debugging or upstreaming to OpenBLAS?

mattip:
The changes in the patch still are valid, although the line numbers have changed a bit. I tried it and the library builds, so it would be nice to put this in for windows, especially in light of the different licensing of quadmath (LGPL) which makes the static linking on windows problematic, see https://github.com/numpy/numpy/issues/27080

mattip:
I rebased off main and updated the line numbers in the patch.

mattip:
The dynamic link command [is failing](https://github.com/MacPython/openblas-libs/actions/runs/10186384794/job/28178044442#step:7:95) since [`$dynamic_libname`](https://github.com/MacPython/openblas-libs/blob/39200966524a0447af5f7edef8df63cb985b2fad/tools/build_gfortran.sh#L13) is empty, somehow the build is not creating the expected `*.dll.a` anymore. [This line in the changeset](https://github.com/MacPython/openblas-libs/pull/85/files#diff-155da1b03b04e0e2d530af5ff2609ab885186864b733400b885573c7dfa9ce70R23) seems relevant, I wonder why it is not creating the import library. I will try to reproduce locally.

mattip:
Sorry, I am not sure I understand. What is the difference between the latest suggestion and the patch?

mattip:
Are you sure? When I changed `FEXTRALIBS` to exlude `-lquadmath`, OpenBLAS no longer linked (on windows), `quadmath_snprintf` was missing. Here is the diff I used:
```
diff --git a/f_check b/f_check
index 93c5962de..ce326b614 100755
--- a/f_check
+++ b/f_check
@@ -380,7 +380,7 @@ if [ -n "$link" ]; then
         case "$flag" in -l*)
             case "$flag" in
                 *ibrary*|*gfortranbegin*|*flangmain*|*frtbegin*|*pathfstart*|\
-                    *crt[0-9]*|*gcc*|*user32*|*kernel32*|*advapi32*|*shell32*|\
+                    *crt[0-9]*|*gcc*|*user32*|*kernel32*|*advapi32*|*shell32*|quadmath*|\
                     -l) ;;
                 *omp*)
                     case "$vendor" in
```

Currently the patch includes `-Wl,--defsym,quadmath_snprintf=snprintf`, but does not modify `f_check`.

mattip:
Thanks, that helps clear up what is going on. 
> If `quadmath_snprintf` is part of the dll at the end can be checked with `-Wl,-Map,output.map`

Cool. I will see if I can make this part of the build script.

mattip:
Adding the symbol redefine `-Wl,--defsym,quadmath_snprintf=snprintf` works on windows-64bit since it links with `-lucrt`. I am not where where the symbol lives in `mingw32`, I cannot use `-lucrt` there.

mattip:
It turns out 32-bit prefixes a `_` to the symbol names, so I had to do `-Wl,--defsym,_quadmath_snprintf=__snprintf_c`. Now CI is passing. In the map file, I see the symbol being remapped:
```
LOAD C:/rtools40/mingw32/bin/../lib/gcc/i686-w64-mingw32/8.3.0/crtbegin.o
                0x6cc37410                _quadmath_snprintf = __snprintf_c

```

matthew-brett:
Should we worry about the issues raised here: https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/3637a7f8-cafd-6eaf-0d75-bbf79daf07c5%40foxvalley.net/#msg36930460 ?

mattip:
If I understand the concerns there, it is that someone may expect the behaviour of `snprintf()` when we are linking to `_snprintf`, and those functions [are different](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-snprintf-snprintf-l-snwprintf-snwprintf-l?view=msvc-140#remarks) around edge cases and return values. The use in libgfortran of `quadmath_snprintf` is [mentioned here](https://github.com/gcc-mirror/gcc/blob/083918a343d6cb9fd28c8b47dd1138220d95c820/libgfortran/io/transfer128.c#L63). I think they mean [this use of snprintf](https://github.com/gcc-mirror/gcc/blob/master/libgfortran/io/write.c#L2247) and another one in that file that is very similar, neither of those checks the return value and neither will run into buffer overruns as far as I can tell: they use very controlled calls to `snprintf`.

mattip:
I would like to merge this and get scipy-openblas wheels built for consideration in the upcoming NumPy 2.1 release, to mitigate issue  https://github.com/numpy/numpy/issues/27080. Any objections?

matthew-brett:
Do we currently trigger uses of the relevant `snprintf` calls in tests?   Just to check if they crash or produce wrong output?

mattip:
> Do we currently trigger uses of the relevant snprintf calls in tests?

No idea. Any thoughts how to check it?

mattip:
Maybe via a LD_PRELOAD stub shared object on linux that exports quadmath_snprintf (since on linux we dynamically link)?

matthew-brett:
Would we be better placed to make a stub `quadmath_snprintf` implementation ourselves, that generates a suitable warning?

mattip:
I am not so concerned that the slight differences between `_snprinf` and the UCRT `snprintf` will cause problems. Only on 32-bits `_snprintf` is used, and NumPy does not even have a `complex256` dtype on windows64, much less on the seldom-used windows32.

mattip:
> Some years ago I replaced `quadmath_snprintf` with abort, but it never hits.

Is that code still around?

matthew-brett:
Sorry to be confusing - my worry was rather that we somehow generate a segfault or similar with this relinking, triggered when some BLAS / LAPACK routine ends up calling `quad_snprintf`, and that we'll only find out when we hit a code path where that happens.   I guess we'd need to find out where in the Fortran code `quad_snprintf` is being called, and work out some way of hitting that code.

mattip:
Ahh. There is some [test code for quadmath](https://people.math.sc.edu/Burkardt/f77_src/gfortran_quadmath/gfortran_quadmath.html). We could 
- get that to work under linux, and use `ltrace -l quadmath` to see if anything there calls `quadmath_snprinf`
- then get that running under mingw, and then use the same link flag to replace the symbol, and make sure tests pass

But I think your analysis when you replaced `quadmath_snprintf` is correct: there is a vanishingly small possibility code will hit that code path. NumPy/SciPy do not use quadmath at all. In order to hit the code path someone would need to write their own code that uses quadmath, but not call any quadmath routines other than `quadmath_snprintf` (since all the other routines are not linked into the dll), build a python library out of it, still use the OpenBLAS dll from NumPy/Scipy, and run that code on windows

matthew-brett:
Thanks for the analysis.  My suggestion before was to write some stub that would catch that very rare case and give some informative error, just in case.

mattip:
OK, let's see how hard it is to add a stub that prints a warning to stderr and calls `snprintf`/`_snprintf`.

rgommers:
Hey all. Unless the stub is like <= 1hr of work _and_ can be finished today, I would strongly suggest that it's way more important to get new wheels up than mitigate some very low-probability event with a better error message. NumPy 2.1.x just branched 2.5 hours ago and if we miss the boat there we both have the potential license concern being unresolved and may cause an extra numpy RC to be needed, delaying the rollout for Python 3.13 support.

mattip:
Sorry, I don't follow. Are you suggesting we replace `quadmath_snprintf` with a segfault or noop? I thought you wanted a warning?

mattip:
> Not to say that this symbol is not exported in every case.
> My conclusion is: replace `quadmath_snprintf`` with a no-op operation.

If I understand correctly, then it really doesn't matter what replaces `quadmath_snprintf`, it will not be called in the current win64/win32 builds of OpenBLAS. So what is the difference between adding a no-op function and using the current code (which replaces it with `snprintf/_snprintf_c`)?

mattip:
Closing, this was merged in the v0.3.28 builds in #178 and backported to the v0.3.27 builds, thanks to @carlkl and @matthew-brett for the original patches and for mentoring the progress.

mattip:
It may be interesting to see how  numpy/numpy-user-dtypes#98 progresses, and if it makes us rethink using `quadmathlib` on windows.

PR #91, opened 2022-11-09
update gfortran-install submodule

mattip:
Use gfortran 11.3 from conda for macos x86_64 ( from MacPython/gfortran-install#11). The latest 0.3.21 macos uploads were mistakenly labeled 10.11, they are actually 10.14 instead.

### DIFF
--- a/.github/workflows/multibuild.yml
+++ b/.github/workflows/multibuild.yml
@@ -53,7 +53,7 @@ jobs:
     env:
       REPO_DIR: OpenBLAS
       OPENBLAS_COMMIT: "v0.3.21"
-      MACOSX_DEPLOYMENT_TARGET: 10.11
+      MACOSX_DEPLOYMENT_TARGET: 10.9
       MB_PYTHON_VERSION: ${{ matrix.python-version }}
       TRAVIS_PYTHON_VERSION: ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


isuruf:
You can change to 10.9 now with https://github.com/MacPython/gfortran-install/pull/11

mattip:
I wonder if somehow the gfortran-install submodule should indicate it supports `MACOSX_DEPLOYMENT_TARGET>=10.9`

mattip:
The artifacts have different suffixes, previously the x86_64 suffix had `gf_1becaaa` in it
```
openblas-v0.3.21-macosx_10_9_x86_64-gf_c469a42.tar.gz
openblas64_-v0.3.21-macosx_10_9_x86_64-gf_c469a42.tar.gz
openblas-v0.3.21-macosx_11_0_arm64-gf_5272328.tar.gz
openblas64_-v0.3.21-macosx_11_0_arm64-gf_5272328.tar.gz
```

isuruf:
That's expected. The gfortran compiler changed and the hash of the compiler is used in the build.

mattip:
right, that is `GFORTRAN_SHA` from the gfortran-install repo. So this will require an update to the numpy openblas_support script in numpy/numpy#22525

mattip:
ppc64le builds are failing, like the previous ones. That is a best-effort attempt to support that platform, so not a blocker.

PR #93, opened 2023-02-08
update to latest OpenBLAS devel, update multibuild

mattip:

PR #94, opened 2023-02-09
actually update the openblas git tag

mattip:
#93 just updated the submodule without updating the tag :frowning_face:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ on:
     branches: [ master ]
 
 env:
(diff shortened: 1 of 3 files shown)


mattip:
The command to get the tag is `$ git describe --tags --abbrev=8` which for the latest HEAD of the `develop` branch reports `v0.3.20-571-g3dec11c6`. This is because the `v0.3.21` tag appears on a different branch, not the `develop` branch. I think merging the branch with the tag back to develop would fix that.
 
cc @martin-frbg

Issue #100, opened 2023-06-15
Upload nightly (weekly) builds to anaconda.org


mattip:
It would be good to reduce the lag between OpenBLAS changes and getting them tried out in SciPy/NumPy. The first step in that would be to upload like NumPy/SciPy/matplotlib do to https://anaconda.org/scientific-python-nightly-wheels/.

xref @honno @steppi

rgommers:
Cc @tupui is one of the admins for that Anaconda bucket - he can help with creating a token that's needed for uploading.

tupui:
The request is approved and we can move forward. Please get in touch (privately) to share the Anaconda handles of the admins for this.

(We are discussing with Anaconda's team right now about our security concerns, etc. So things should hopefully improve.)

mattip:
The scheduled linux runs [look good](https://anaconda.org/multibuild-wheels-staging/openblas-libs/files). For some reason the macos ones apparently did not pick up the `NIGHTLY` environment variable, and built 0.3.23 tarballs.

mattip:
Reopening to change the frequency to once-a-week, and to add in the travis builds

mattip:
Additionally, we should move the [upload target](https://github.com/MacPython/openblas-libs/blob/a64e515f2db3d226c1bfa603050b99ee517d18e3/.github/workflows/multibuild.yml#L139) from https://anaconda.org/multibuild-wheels-staging/openblas-libs/ to https://anaconda.org/scientific-python-nightly-wheels/openblas-libs

steppi:
> Additionally, we should move the [upload target](https://github.com/MacPython/openblas-libs/blob/a64e515f2db3d226c1bfa603050b99ee517d18e3/.github/workflows/multibuild.yml#L139) from https://anaconda.org/multibuild-wheels-staging/openblas-libs/ to https://anaconda.org/scientific-python-nightly-wheels/openblas-libs

does just mean replacing

```
--no-progress -u multibuild-wheels-staging
```

with

```
--no-progress -u scientific-python-nightly-wheels
```
?

mattip:
> does just mean replacing

yes, but the proper token is needed as well. I think @honno has permissions needed to make this happen

steppi:
> > does just mean replacing
> 
> yes, but the proper token is needed as well. I think @honno has permissions needed to make this happen

Thanks. Got it.

tupui:
[...]

PR #103, opened 2023-07-04
Nightly linux builds

honno:
Resolves https://github.com/MacPython/openblas-libs/issues/100. 

As suggested by @MattiP, I've tackled this by having the nightly builds be uploaded to the same filenames with "HEAD" where the version would be. This is a crude approach compared to how other nightlies work (see [`scipy-wheels-nightly`](https://anaconda.org/scipy-wheels-nightly)), but should be fine for the purposes of testing NumPy and SciPy against OpenBLAS nightlies.

<details><summary>Example filenames</summary>

```
openblas-HEAD-musllinux_1_1_x86_64.tar.gz
openblas-HEAD-manylinux2014_x86_64.tar.gz
openblas64_-HEAD-musllinux_1_1_x86_64.tar.gz
openblas64_-HEAD-manylinux2014_x86_64.tar.gz
openblas64_-HEAD-manylinux2010_x86_64.tar.gz
openblas-HEAD-manylinux2014_i686.tar.gz
openblas-HEAD-manylinux2010_x86_64.tar.gz
openblas-HEAD-manylinux2010_i686.tar.gz
```

</details>

This PR is only generating linux builds (basically copy-pasted the `multibuild.yml` workflow). Some linux builds fail, which I can have a look at but might be a cakewalk to resolve for you folks :sweat_smile: As the new workflow runs on cron rather than push, you can look at https://github.com/honno/openblas-libs/pull/2 (test PR I made). From the latest [Actions run](https://github.com/honno/openblas-libs/actions/runs/5457702311), the following jobs are currently failing:

```
nightly multibuild / build (ubuntu-latest, x64, i686, 1)
nightly multibuild / build (ubuntu-latest, x64, i686, 1, 2010)
nightly multibuild / build (ubuntu-latest, x64, i686, 1, 2014)
```

</details>

Also note I removed Mac and Windows builds just so I could push a MVP on nightlies out the door, but I could re-instate Mac builds at least as they already work with `multibuild.yml` and thus should with `nightly_multibuild.yml` (although when I tried before they failed too).

cc @steppi

### DIFF
--- a/travis-ci/docker_build_wrap.sh
+++ b/travis-ci/docker_build_wrap.sh
@@ -7,4 +7,4 @@ set -e
 # Change into root directory of repo
 cd /io
 source travis-ci/build_steps.sh
-do_build_lib "$PLAT" "" "$INTERFACE64"
+do_build_lib "$PLAT" "" "$INTERFACE64" "$NIGHTLY"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,5 +82,6 @@ jobs:
         env:
          OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN: ${{ secrets.MULTIBUILD_WHEELS_STAGING_ACCESS }}
         run: |
-          conda install anaconda-client
+          # Pin urllib3<2 due to github.com/Anaconda-Platform/anaconda-client/issues/654
+          conda install "urllib3<2" anaconda-client
           & $env:BASH_PATH -lc tools/upload_to_anaconda_staging.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,9 +2,9 @@ name: Win
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
   pull_request:
-    branches: [ master ]
+    branches: [ main ]
 
 env:
   OPENBLAS_COMMIT: "v0.3.23"
--- a/.github/workflows/multibuild.yml
+++ b/.github/workflows/multibuild.yml
@@ -2,9 +2,9 @@ name: multibuild
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
   pull_request:
-    branches: [ master ]
+    branches: [ main ]
 
 jobs:
   build:
--- a/.github/workflows/multibuild.yml
+++ b/.github/workflows/multibuild.yml
@@ -5,6 +5,9 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+  workflow_dispatch: null
+  schedule:
+    - cron: '0 0 * * *'
 
 jobs:
   build:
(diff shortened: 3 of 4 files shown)


travis-ci/docker_build_wrap.sh
```diff
@@ -7,4 +7,4 @@ set -e
 # Change into root directory of repo
 cd /io
 source travis-ci/build_steps.sh
-do_build_lib "$PLAT" "" "$INTERFACE64"
+do_build_lib "$PLAT" "" "$INTERFACE64" "$NIGHTLY"
```
mattip, line 10: This is the one used by the docker image for linux

.github/workflows/nightly_multibuild.yml
```diff
@@ -0,0 +1,115 @@
...
+      with:
+        submodules: recursive
+        fetch-depth: 0
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Set extra env
+      run: |
+        git ls-remote https://github.com/xianyi/OpenBLAS HEAD | sed -e "s/\s*HEAD$//" | xargs -I {} echo "OPENBLAS_COMMIT={}" >> $GITHUB_ENV;
```
mattip, line 59: Couldn't you use `OPENBLAS_COMMIT=HEAD` or `OPENBLAS_COMMIT=develop` instead? Then rather than add a `NIGHTLY` env variable, you could check `$OPENBLAS_COMMIT == "HEAD"`.

mattip (commented):
Looks good. With an eye to merging this with the other build, we could do something like this to identify a cron run or a manually-triggered CI run
```
env:
  NIGHTLY: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
```

and use that to override other settings like the tarball name and OPENBLAS_COMMIT

honno:
This PR now starts from https://github.com/MacPython/openblas-libs/pull/104, as it'll need it anyway to work.

> With an eye to merging this with the other build, we could do something like this to identify a cron run or a manually-triggered CI run
> 
> ```
> env:
>   NIGHTLY: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
> ```
> 
> and use that to override other settings like the tarball name and OPENBLAS_COMMIT

> Couldn't you use `OPENBLAS_COMMIT=HEAD` or `OPENBLAS_COMMIT=develop` instead? Then rather than add a `NIGHTLY` env variable, you could check `$OPENBLAS_COMMIT == "HEAD"`.

So I've mangled nightly builds into `multibuild.yml`, using your suggested method of defining a `NIGHTLY` env. It seems to work nicely enough, but I don't have a sense for these things.

Unfortunately I couldn't figure out why bash scripts wouldn't recognise env variables on GitHub Actions, where they would when run locally. The pass through of `NIGHTLY` is a bit noisy (i.e. https://github.com/MacPython/openblas-libs/pull/103#discussion_r1252717202 /seems/ necessary as `dcker_build_wrap.sh` is used in `build_steps.sh::build_lib`, so did want to just check the env variable directly when defining [`versions` in `build_steps.sh::do_build_lib`](https://github.com/MacPython/openblas-libs/blob/062f1f6e3ff102287bae00833e9ea245a9f68d61/travis-ci/build_steps.sh#L169) instead:

```diff
-     local version=$(cd OpenBLAS && git describe --tags --abbrev=8)
+     if [ "$NIGHTLY" = "true" ]; then
+         local version="HEAD"
+     else
+         local version=$(cd OpenBLAS && git describe --tags --abbrev=8)
+     fi
```

mattip:
Let's give it a shot. I can do a manual workflow run once this is merged.

PR #106, opened 2023-07-13
Change version for nightly on anaconda to HEAD

steppi:
This PR resolves an issue that caused the path to nightly openblas-libs builds to contain the openblas version number instead of HEAD. Currently paths look like

```
https://anaconda.org/multibuild-wheels-staging/openblas-libs/v0.3.23/download/openblas64_-HEAD-manylinux2014_x86_64.tar.gz
```

instead of

```
https://anaconda.org/multibuild-wheels-staging/openblas-libs/HEAD/download/openblas64_-HEAD-manylinux2014_x86_64.tar.gz
```

@mattip identified that the issue is in the line `-v "$(cd OpenBLAS && git describe --tags --abbrev=8)"` that is in both `multibuild.yml` and `build_steps.sh``.

I've fixed both files. `multibuild.yml` was straightforward to fix because the script to upload to anaconda is aware of whether `NIGHTLY` is true. For `multibuild.yml` the function to upload to anaconda currently just searches for files matching the pattern `openblas*.tar.gz`. I've separated this into two function, one which matches `openblas*HEAD*.tar.gz`, and the other which matches all strings of the form `openblas*.tar.gz` which do not contain `"HEAD"`.

### DIFF
--- a/.github/workflows/multibuild.yml
+++ b/.github/workflows/multibuild.yml
@@ -138,12 +138,13 @@ jobs:
               -t file -p "openblas-libs"
               -d "OpenBLAS for multibuild wheels"
               -s "OpenBLAS for multibuild wheels"
-              -v "$(cd OpenBLAS && git describe --tags --abbrev=8)"
             )
             if [[ "$NIGHTLY" = "true" ]]; then
               args+=(--force)
+              args+=(-v "HEAD")
             else
               args+=(--skip)
+              args+=(-v "$(cd OpenBLAS && git describe --tags --abbrev=8)")
             fi
             anaconda "${args[@]}" libs/openblas*.tar.gz
         fi


steppi:
cc @mattip @honno

tools/upload_to_anaconda_posix.sh
```diff
@@ -7,4 +7,5 @@ echo $(python -c "import sys; print(sys.version)")
 sudo chmod -R a+w /home/travis/.cache
 pip install -q git+https://github.com/Anaconda-Platform/anaconda-client@1.8.0
 upload_to_anaconda
+upload_nightlies_to_anaconda
```
mattip, line 10: This will run both functions. I think it would be better to use the `$NIGHTLY` environment variable like you did above. Does it exist in the environment at this point?

mattip: ahh, this is used by the travis build (aarch64, s390x, ppc64le) which has not yet been converted to use `NIGHTLY` and `HEAD`, so it is not needed yet.

steppi: > This will run both functions. I think it would be better to use the `$NIGHTLY` environment variable like you did above. Does it exist in the environment at this point?

Right, for the travis build, I don't think `$NIGHTLY` existed at this point. `upload_to_anaconda`  just finds files to upload by matching the pattern `libs/openblas*.tar.gz`. Running both functions as I implemented them would have lead to the same basic ibehavior, except that nightlies would have had `HEAD` in the right part of their urls instead of the version.



> ahh, this is used by the travis build (aarch64, s390x, ppc64le) which has not yet been converted to use `NIGHTLY` and `HEAD`, so it is not needed yet.

OK, I'll just get rid of these new steps in the travis build.

mattip:
I think this is only relevant after merge, so I will merge it before CI finishes. Thanks @steppi

PR #107, opened 2023-07-14
INTERFACE64 cannot be blank

mattip:
since the code to use nightly added another parameter to `build_lib`, the INTERFACE64 parameter cannot be "", it must be "0". Otherwise, in the call to `build_lib "$PLAT" "$INTERFACE64" "1", the `1` is interpreted as the `INTERFACE64` argument.


mattip:
cc @honno

mattip:
Self merging to clear build errors

PR #109, opened 2023-07-16
use 'develop' for nightly builds

### LINKED ISSUE #108: HEAD build is not building the correct commit
~Any build that is not directly from a tag should have `dev` in the version or some other way to get more appropriate information. Currently, `openblas_get_config()` returns something like `OpenBLAS 0.3.23  USE64BITINT DYNAMIC_ARCH NO_AFFINITY Zen MAX_THREADS=64` even for builds that are not exactly `0.3.23`. It would be nice if it were `OpenBLAS 0.3.24-dev  ...`. @martin-frbg is there a way to do that in the `make` command`?~

OpenBLAS is not the problem, we are not properly checking out the commit we want

mattip:
Closes #108

### DIFF
--- a/.github/workflows/multibuild.yml
+++ b/.github/workflows/multibuild.yml
@@ -108,12 +108,15 @@ jobs:
         source travis-ci/build_steps.sh
         echo "------ BEFORE BUILD ---------"
         before_build
-        echo "------ CLEAN CODE --------"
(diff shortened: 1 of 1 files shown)


mattip:
No need to wait for CI, since this will only activate on the nightly builds

PR #110, opened 2023-07-17
use 8 characters in git describe on windows

mattip:
The latest windows builds are using 9 characters instead of 8. Maybe something changed with git? In any case, using `--abbrev 8` should fix the problem

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: install-rtools
         run: |
           # rtools 42+ does not support 32 bits builds.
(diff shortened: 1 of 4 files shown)


mattip:
The other uses of `git describe` (on github actions and travis for macos and linux) all use `--abbrev 8`. The [documentation for config.abbrev](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreabbrev) says
> If unspecified or set to "auto", an appropriate value is computed based on the approximate number of packed objects in your repository, which hopefully is enough for abbreviated object names to stay unique for some time. 

So the OpenBLAS repo must have crossed a number of commits that triggered a change from 8 to 9.

mattip:
actions/checkout@v2 was emitting a warning, update to v3

mattip:
Not waiting for CI: the important things passed on the previous commit, and the HEAD commit passed the "checkout" stage.

PR #111, opened 2023-07-18
Upload to `scientific-python-nightly-wheels`, change nightly frequency to weekly

honno:
Resolves outstanding issues from https://github.com/MacPython/openblas-libs/issues/100#issuecomment-1630679025:
* Changes the frequency of the nightly linux/mac builds to once-a-week... so they're now really weeklies!
* Moves over ALL builds from [`multibuild-wheels-staging/openblas-libs`](https://anaconda.org/multibuild-wheels-staging/openblas-libs) to [`scientific-python-nightly-wheels/openblas-libs`](https://anaconda.org/scientific-python-nightly-wheels/openblas-libs)

cc @mattip @steppi

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,7 +80,7 @@ jobs:
 
       - name: Upload
         env:
-          ANACONDA_SCIENTIFIC_PYTHON_UPLOAD: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
+         ANACONDA_SCIENTIFIC_PYTHON_UPLOAD: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
         run: |
           # Pin urllib3<2 due to github.com/Anaconda-Platform/anaconda-client/issues/654
           conda install "urllib3<2" anaconda-client
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,7 +80,7 @@ jobs:
 
       - name: Upload
         env:
-         OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN: ${{ secrets.MULTIBUILD_WHEELS_STAGING_ACCESS }}
+         OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
         run: |
           # Pin urllib3<2 due to github.com/Anaconda-Platform/anaconda-client/issues/654
           conda install "urllib3<2" anaconda-client
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,7 +80,7 @@ jobs:
 
       - name: Upload
         env:
-         OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
+          ANACONDA_SCIENTIFIC_PYTHON_UPLOAD: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
         run: |
           # Pin urllib3<2 due to github.com/Anaconda-Platform/anaconda-client/issues/654
           conda install "urllib3<2" anaconda-client
--- a/tools/upload_to_anaconda_staging.sh
+++ b/tools/upload_to_anaconda_staging.sh
@@ -16,7 +16,7 @@ else
     ls -lh builds/openblas*.zip
 
     anaconda -t $OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN upload \
-            --no-progress --force -u multibuild-wheels-staging \
+            --no-progress --force -u scientific-python-nightly-wheels \
             -t file -p "openblas-libs" -v "$VERSION" \
             -d "OpenBLAS for multibuild wheels" \
             -s "OpenBLAS for multibuild wheels" \
--- a/tools/upload_to_anaconda_staging.sh
+++ b/tools/upload_to_anaconda_staging.sh
@@ -9,13 +9,13 @@ pushd OpenBLAS
 VERSION=$(git describe --tags --abbrev=8)
 popd
 
-if [ "$OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN" == "" ]; then
-    echo "OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN is not defined: skipping."
+if [ "$ANACONDA_SCIENTIFIC_PYTHON_UPLOAD" == "" ]; then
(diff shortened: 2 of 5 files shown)


honno:
@martin-frbg Could you just confirm you're happy with weekly builds being uploaded to [`scientific-python-nightly-wheels`](https://anaconda.org/scientific-python-nightly-wheels)? I ask as its helpful to have public acknowledgement from an OpenBLAS maintainer. cc @tupui

tupui:
Thank you for the confirmation.

For the advertising, you can add a warning on Anaconda's page if you think the name would not be clear enough for users.

tools/upload_to_anaconda_staging.sh
```diff
@@ -16,7 +16,7 @@ else
     ls -lh builds/openblas*.zip
 
     anaconda -t $OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN upload \
```
mattip, line 18: this needs to change to the new name `ANACONDA_SCIENTIFIC_PYTHON_UPLOAD`, both here and in the travis run

.github/workflows/build.yml
```diff
@@ -80,7 +80,7 @@ jobs:
 
       - name: Upload
         env:
-         OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN: ${{ secrets.MULTIBUILD_WHEELS_STAGING_ACCESS }}
+         OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
```
mattip, line 83: Since the `upload_to_anaconda_staging.sh` script looks for an environment variable, you need to sync the env name with this stanza. So either this should be
```
 ANACONDA_SCIENTIFIC_PYTHON_UPLOAD: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
```

or the name in `tools/upload_to_anaconda_staging.sh` should remain unchanged.

mattip: Just to be clear: the `tools/upload_to_anaconda_staging.sh` script is used only for this windows CI job. The others use `travis-ci/build_steps.sh`.

.github/workflows/build.yml
```diff
@@ -25,6 +25,8 @@ jobs:
             INTERFACE64: 1
       fail-fast: false
     runs-on: ${{ matrix.os }}
+    env:
+      ANACONDA_SCIENTIFIC_PYTHON_UPLOAD: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
```
mattip, line 29: This should not be in the global environment, rather in the Upload-specific environment below

.github/workflows/build.yml
```diff
@@ -80,7 +82,7 @@ jobs:
 
       - name: Upload
         env:
-         OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN: ${{ secrets.MULTIBUILD_WHEELS_STAGING_ACCESS }}
+         OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
```
mattip, line 85:
```suggestion
         ANACONDA_SCIENTIFIC_PYTHON_UPLOAD: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
```

.github/workflows/build.yml
honno: Completely missed that :facepalm:

.github/workflows/build.yml
```diff
@@ -80,7 +80,7 @@ jobs:
 
       - name: Upload
         env:
-         OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN: ${{ secrets.MULTIBUILD_WHEELS_STAGING_ACCESS }}
+          ANACONDA_SCIENTIFIC_PYTHON_UPLOAD: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
```
mattip, line 83: one too many spaces...

.github/workflows/build.yml
```diff
@@ -80,7 +80,7 @@ jobs:
 
       - name: Upload
         env:
-         OPENBLAS_LIBS_STAGING_UPLOAD_TOKEN: ${{ secrets.MULTIBUILD_WHEELS_STAGING_ACCESS }}
+         ANACONDA_SCIENTIFIC_PYTHON_UPLOAD: ${{ secrets.ANACONDA_SCIENTIFIC_PYTHON_UPLOAD }}
```
mattip, line 83: heh, maybe you were right and now, like before, there is one space too little. I was not looking carefully. Well, lets see if it works.

mattip:
Let's give it a shot...

rgommers:
I see only new Windows builds at https://anaconda.org/scientific-python-nightly-wheels/openblas-libs/files, and there's a HEAD vs. xxx-commit-hash naming difference. I guess only GHA triggered and not TravisCI? Just checking if it's safe to merge the connected SciPy PR.

mattip:
Since these are tarballs not wheels, I think it is not worth the trouble to separate the two. For wheels, it is important not to mix them.

However, the tarballs needed currently by SciPy are not here, so for the mean time SciPy non-nightly builds should continue to download from the old site. Once SciPy updates to a newer OpenBLAS tag, it can download from the new site.

steppi:
> Since these are tarballs not wheels, I think it is not worth the trouble to separate the two. For wheels, it is important not to mix them.
> 
> However, the tarballs needed currently by SciPy are not here, so for the mean time SciPy non-nightly builds should continue to download from the old site. Once SciPy updates to a newer OpenBLAS tag, it can download from the new site.

Got it, thanks. Of course, they're not wheels, I should pay better attention.

mattip:
The travis upload is working now, so at least NumPy can use the new site for the non-nightly builds.

mattip:
I started a workflow. It correctly got the [`Nightly=true`](https://github.com/MacPython/openblas-libs/actions/runs/5653509777/job/15314803357#step:4:15) setting.

rgommers:
It looks like everything uploaded to https://anaconda.org/scientific-python-nightly-wheels/openblas-libs/files. 



> and there's a HEAD vs. xxx-commit-hash naming difference.

Still not sure if the three Windows builds should be `xxx-commit-hash` vs. the rest using `HEAD`, but we're not using Windows nightlies yet, so I'll go revisit the SciPy PR now.

mattip:
We are using one bucket for both the OpenBLAS-version-specific tarballs (ones with `xxx-commit-hash`) and the OpenBLAS-HEAD tarballs. So there will be a mixture of HEAD and `xxx-commit-hash` tarballs here. Nightlies (now actually "weeklies") are produced for `manylinux-x86*` only, so there should be fewer of those. There are 23 tarballs with the `v0.3.23-246-g3d31191b` tag, which is as it should be.

rgommers:
Ah, got it, thank you!

PR #116, opened 2023-09-26
Continue wheel building

mattip:
- Update the wheel for the current build to 0.3.23.293.1
- create nightly wheels on linux only
- upload all the wheels to scientific-python-nightly-wheels
- move files from travis-ci to tools

And do some cleanups from review of #87
- properly package license files like on NumPy
- remove wrong pkgconfig files from the wheel's lib directory


mattip:
Merged. Let's see if the upload works...

PR #121, opened 2023-09-28
use +e to execute the upload script

mattip:


mattip:
Trying the `cmd && true` trick didn't work. Merging since the change is only active on `main`

PR #122, opened 2023-10-01
fix nightly version parsing

mattip:


mattip:
merging since this only affects scheduled or manually triggered builds

PR #125, opened 2023-10-29
Add symbolprefix=scipy

mattip:
It turns out we need to add a symbol prefix in order to use the wheels, otherwise there are clashes with other builds of OpenBLAS and/or other BLAS implementations. This adds a `SYMBOLPREFIX=scipy` to these builds. Also
- update the OpenBLAS hash to latest HEAD
- update the OpenBLAS and multibuild subrepos (not necessarily required)

### DIFF
--- a/.github/workflows/posix.yml
+++ b/.github/workflows/posix.yml
@@ -69,6 +69,7 @@ jobs:
     - name: Set extra env
       run: |
         # if [ "macos-11" == "${{ matrix.os }}" ]; then
+        #   Use xcrun --sdk macosx --show-sdk-path instead of hardcoding the path
         #   echo "LDFLAGS=-L/Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/usr/lib" >> $GITHUB_ENV;
         #   echo "LIBRARY_PATH=-L/Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/usr/lib" >> $GITHUB_ENV;
         # else
--- a/.github/workflows/posix.yml
+++ b/.github/workflows/posix.yml
@@ -46,7 +46,7 @@ jobs:
             INTERFACE64: '1'
     env:
       REPO_DIR: OpenBLAS
-      OPENBLAS_COMMIT: "c2f4bdb"
+      OPENBLAS_COMMIT: "9d425a5f"
       NIGHTLY: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
       MACOSX_DEPLOYMENT_TARGET: 10.9
       MB_PYTHON_VERSION: ${{ matrix.python-version }}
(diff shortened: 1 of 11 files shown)


mattip:
This is ready if anyone wants to review. I don't think we need to change the wheel names for the changed function names. This is a new version `0.3.24.95` and the changed prefix requires no external code changes.

.github/workflows/posix.yml
```diff
@@ -68,17 +68,14 @@ jobs:
...
-        if [ "macos-11" == "${{ matrix.os }}" ]; then
-          echo "TRAVIS_OS_NAME=osx" >> $GITHUB_ENV;
-          echo "LDFLAGS=-L/Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/usr/lib" >> $GITHUB_ENV;
-          echo "LIBRARY_PATH=-L/Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/usr/lib" >> $GITHUB_ENV;
-        else
-          echo "TRAVIS_OS_NAME=${{ matrix.os }}" >> $GITHUB_ENV;
-        fi
+        # if [ "macos-11" == "${{ matrix.os }}" ]; then
+        #   echo "LDFLAGS=-L/Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/usr/lib" >> $GITHUB_ENV;
+        #   echo "LIBRARY_PATH=-L/Library/Developer/CommandLineTools/SDKs/MacOSX12.1.sdk/usr/lib" >> $GITHUB_ENV;
```
rgommers, line 73: I'll note that if this is needed again in the future or changes, there's a better way than hardcoding it. The path can be obtained from `xcrun --sdk macosx --show-sdk-path`. I just managed to get rid of this hardcoding in numpy CI.

mattip: I added a comment in the commit to add the BLAS_SYMBOL_PREFIX definition to the pkgconfig file

mattip:
Currently if you use the ILP64 builds, you have to add `-Duse-ilp64=true` which ends up defining `OPENBLAS_ILP64_NAMING_SCHEME`. We have cflags in the pkgconfig file, so I added the macro there. Once use of these newer wheels is normalized we can remove the meson flag.

rgommers:
We can't in general, because there are libraries that have both LP64 and ILP64 symbols. True right now for at least MKL and Accelerate, and may also be true for OpenBLAS sometime next year (after reference BLAS/LAPACK adopts it). It has to be a user opt-in. It may be okay for the opt-in to be automatic for `scipy-openblas`, less sure of that.

mattip:
> It may be okay for the opt-in to be automatic for scipy-openblas, less sure of that.

It worked locally, with the latest commit I can do
```
PKG_CONFIG_PATH=$PWD/.openblas
pip install scipy-openblas64
spin build --with-scipy-openblas=64  # without -- -Duse-ilp64=true
```

PR #128, opened 2024-01-19
create a single shared object libscipy_openblas64_.so or libscipy_openblas.so

mattip:
Related to numpy/numpy#25505

Instead of building a set of shared objects and creating symlinks to them, create only a single shared object. Use a patch to OpenBLAS.

### DIFF
--- a/.github/workflows/posix.yml
+++ b/.github/workflows/posix.yml
@@ -46,7 +46,7 @@ jobs:
             INTERFACE64: '1'
     env:
       REPO_DIR: OpenBLAS
-      OPENBLAS_COMMIT: "9d425a5f"
+      OPENBLAS_COMMIT: "v0.3.26"
       NIGHTLY: ${{ github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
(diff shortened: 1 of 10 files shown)


mattip:
There are some downstream projects using the tarball. here is [a github search](https://github.com/search?q=anaconda.org%2Fscientific-python-nightly-wheels%2Fopenblas-libs&type=code). This PR changes the shared objects in the tarball, I wonder if it will break their build. I asked on on the jdtuc/fdasrsf_python project.

rgommers:
They all have copied the numpy/scipy script which pins to an exact hash, so I wouldn't worry about it. This repo is "use at your own risk" anyway.

mattip:
This works locally, even with the older tarballs. But it requires one change from numpy/numpy#25505: [this fix](https://github.com/numpy/numpy/pull/25505/files#diff-d4017f02419a9799bc2d59ae04399d2c45b9855443bb4a5ff7544ff76b040258R246) for when scipy-openblas is detected by meson. So in order not to override the current tarball, which is used by numpy builds, I will bump the OpenBLAS version, which is overdue anyway.

PR #136, opened 2024-02-05
audtiwheel6 fixed an issue, remove the work-around

mattip:
- [ ] I updated the version in pyproject.toml and made sure it matches `git describe --tags --abbrev=8` in OpenBLAS at the `OPENBLAS_COMMIT`


mattip:
It seems auditwheel 6 did not fix the problem
xref https://github.com/pypa/auditwheel/issues/451

PR #141, opened 2024-02-25
use new LIBNAMEPREFIX, LIBNAMESUFFIS, FIXED_LIBNAME options instead of patching

mattip:
- [ ] I updated the version in pyproject.toml and made sure it matches `git describe --tags --abbrev=8` in OpenBLAS at the `OPENBLAS_COMMIT`


mattip:
No idea what is wrong with travis, it seems the builds are not progressing even though the build script has not significantly changed.

Issue #142, opened 2024-02-28
the SONAME in the linux shared object is wrong


mattip:
After #141 and removing the local patch, the wheel seems to be produced correctly but the SONAME in the shared object (on linux at least) is wrong: `libscipy_scipy_openblas.so`.

mattip:
Also, https://github.com/OpenMathLib/OpenBLAS/pull/4534 will require fixes on windows for the latest nightly build: now both `SYMBOLPREFIX` and `LIBNAMEPREFIX` (and correspondingly `SUFFIX`) should be defined.

PR #143, opened 2024-02-28
update to latest HEAD, consistently use LIBNAME

mattip:


mattip:
Closes #142

PR #146, opened 2024-03-11
remove wheel build on travis CI

mattip:
This is a PR to build older-style v0.3.26 tarballs and upload them to https://anaconda.org/multibuild-wheels-staging/openblas-libs/


mattip:
I added an appropriate `scipy` branch-limited secret to the travis-ci configuration, so this should upload aarch64, ppc64le, and s390x tarballs after merging

PR #151, opened 2024-04-10
V0.3.27.1

mattip:


mattip:
SONAME was `libscipy_openblas64_64_.so`, is now `libscipy_openblas64_.so`.

PR #160, opened 2024-05-12
patch to fix test warning

mattip:


mattip:
Fixed upstream in https://github.com/OpenMathLib/OpenBLAS/pull/4690

PR #165, opened 2024-07-21
use macos-13 instead of macos-11

mattip:
Change the gitub runner to macos-13 since macos-11 is deprecated.

### DIFF
--- a/.github/workflows/posix.yml
+++ b/.github/workflows/posix.yml
@@ -102,11 +102,11 @@ jobs:
           build_lib "$PLAT" "$INTERFACE64" "0"
         fi
 
-    - name: Setup tmate session
-      if: ${{ failure() }}
-      uses: mxschmitt/action-tmate@v3
-      with:
-        limit-access-to-actor: true
+    # - name: Setup tmate session
+    #   if: ${{ failure() }}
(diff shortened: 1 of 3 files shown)


mattip:
The macos-x86_64 builds are failing with [`dyld[97241]: missing symbol called`](https://github.com/MacPython/openblas-libs/actions/runs/10027868632/job/27714083146#step:8:5833). I think this is when using one of the tools: gfortran, the linker, ... Anyone seen this when moving from macos-11 to macos-13?

mattip:
Ahh, it seems `/usr/local/lib` is not added to the search path on `x86_64`, so the gfortran runtime shared objects are not found. I wonder if the `x86_64` gfortran install should be [soft-linking](https://github.com/MacPython/gfortran-install/blob/3dd38d9ce78b3890598cb0eff18a7bec50c06f5e/gfortran_utils.sh#L127) to a different path? @matthew-brett ideas?  The apple documentation suggests `/usr/local/lib` should be in `DYLD_FALLBACK_LIBRARY_PATH`, but maybe it is not when using `x86_64` on macos 13?

mattip:
This was the best I found, apparently /usr/local/lib is no longer searched https://forums.developer.apple.com/forums/thread/737920

PR #166, opened 2024-07-25
use DYNAMIC_LIST to limit the number of kernels compiled into the wheel

### LINKED ISSUE #144: Shrink x86_64 blas library size
As @martin-frbg [says in a comment elsewhere](https://github.com/numpy/numpy/issues/24905#issuecomment-1981062320)

> Given that there never were Macs with AMD processors, or (AFAIK) with AVX-512, you could reduce the size of your library build by adding DYNAMIC_LIST="CORE2 NEHALEM SANDYBRIDGE HASWELL", removing the dedicated BLAS kernels for 10 irrelevant cpus (if you don't already do this). Probably no longer worth it for you, but I guess I should add this as the default for x86_64 builds with OSNAME=Darwin...

mattip:
Closes #144. Based on the analysis done in that issue.

### DIFF
--- a/tools/build_steps.sh
+++ b/tools/build_steps.sh
@@ -142,9 +142,9 @@ function do_build_lib {
         Linux-aarch64)
             local bitness=64
             local target=ARMV8
-            # temporarily limit the kernels for travis until there is data
+            # temporarily limit the kernels for aarch64 until there is data
             # see https://github.com/MacPython/openblas-libs/issues/170
-            if [ "$MB_ML_LIBC" == "musllinux" ]; then
+            if [ "$MB_ML_LIBC" == "musllinux" -a "$MB_ML_VER" == "_1_1" ]; then
                 local dynamic_list="ARMV8 CORTEXA57 THUNDERX"
             else
                 local dynamic_list="ARMV8 CORTEXA57 NEOVERSEV1 THUNDERX"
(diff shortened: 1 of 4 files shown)


mattip:
Rebased off main and tried limiting the aarch64 builds to see if that clears the Travis error.

tools/build_steps.sh
```diff
@@ -125,32 +125,38 @@ function do_build_lib {
...
+            local target_flags=PRESCOTT
+            local dynamic_list="PRESCOTT NEHALEM SANDYBRIDGE HASWELL"
             ;;
         Linux-aarch64)
             local bitness=64
-            local target_flags="TARGET=ARMV8"
+            local target=ARMV8
+            # temporarily limit the kernels for travis until there is data
+            # see https://github.com/MacPython/openblas-libs/issues/170
+            local dynamic_list="ARMV8 CORTEXA57 THUNDERX"
```
mattip, line 148: If the travis build works, should add back at least `NEOVERSEV1`

mattip:
Rebased off #169 which fixed the travis error by updating to latest HEAD.

mattip:
musllinux_1_1_aarch64 build is failing since it only has gcc version 9.3, which does not have `arm64_sve.h` so the `NEOVERSEV1` build fails. Another reason to move to musllinux_1_2_aarch64 which has gcc 13.2.1, see numpy/numpy#27066

mattip:
Let's limit this to x86_64 for now, we can revisit the aarch64 size after the NumPy 2.1 release.

mattip:
Replaced by #179. Closing. We can try to shrink aarch64 in a diffferent PR/issue.

PR #168, opened 2024-07-26
update to v0.3.27-341-g09298658 for NAN fix

mattip:


mattip:
I don't know why travis CI is timing out. In the past merging to main has worked, let's see if this time it works too

PR #169, opened 2024-07-26
fix CI

mattip:
Only change CI


mattip:
Inside the aarch64 docker build, musllinux successfully runs to completion but glibc fails to build.  
```
../kernel/arm64/gemv_n_sve.c:43:18: error: ACLE function ‘svcntw’ requires ISA extension ‘sve’
    43 | #define SV_COUNT svcntw
       |                  ^~~~~~
 ../kernel/arm64/gemv_n_sve.c:62:25: note: in expansion of macro ‘SV_COUNT’
    62 |     uint64_t sve_size = SV_COUNT();
       |                         ^~~~~~~~
 ../kernel/arm64/gemv_n_sve.c:43:18: note: you can enable ‘sve’ using the command-line \
  option ‘-march’, or by using the ‘target’ attribute or pragma
```

mattip:
Reproducing locally: the A64FX kernel compilation does not set `-march` at all, and since the native architecture does not support SVE, the compilation fails. Strangely, the other SVE kernel builds do work.

mattip:
It seems building inside a aarch64 docker is too slow and fraught with problems. I will have to work out how to convince multibuild to cross-compile.

rgommers:
Perhaps easier to move it to Cirrus CI?

mattip:
> Perhaps easier to move it to Cirrus CI?

I would need access to an account.

rgommers:
> I would need access to an account.

The whole MacPython org is mostly dormant, only a few projects haven't moved their wheel builds away from it yet. Maybe you can simply get org-level access now. @matthew-brett?

mattip:
Getting closer. The build and tests of OpenBLAS succeed, now there is something off about building the wheel out of the artifacts.

mattip:
In the end, aarch64 was failing but because of ... reasons ... stdout was not displayed on travisCI. Adding the ability to build in github CI allowed displaying the error, which led to various fixes including OpenMathLib/OpenBLAS#4826 for building newer aarch64 kernels with gcc 10.4, but using that CI takes over 3 hours, so I moved it back to travis.

PR #177, opened 2024-08-08
Use DYNAMIC_LIST to shrink the shared object on x86_64 and i686

mattip:

I did not add a macos-x86_64 alternative, that platform is in maintenance mode as more people move to macos-arm64

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "scipy-openblas64"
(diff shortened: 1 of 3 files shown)


rgommers:
Thanks Matti!

> I did not add a macos-x86_64 alternative, that platform is in maintenance mode as more people move to macos-arm64

That sounds like a good/pragmatic choice to me.

mattip:
No need to wait for aarch64, the PR does not affect it. Merging.

Issue #175, opened 2024-08-08
Shrink the size of OpenBLAS DLLs on Windows


carlkl:
Shrinking the size of OpenBLAS binary size can be done with several ways:

1. Use the same OpenBLAS DLL for numpy as well for scipy
2. use of DYNAMIC_LIST to reduce the number of targets
3. strip DLL

(1) has the greatest impact on the overall size of a python installation as well as on the memory consumption of a python process. There is no good reason to keep two dedicated OpenBLAS binaries in a process.  
There are two ways to accomplish this: a) scipy could use the OpenBLAS DLL from numpy or b) numpy as well as scipy both depend on a dedicated OpenBLAS wheel with a OpenBLAS.dll included. (b) has the advantage to allow for easy monkey-patching the OpenBLAS DLL, i.e. with less more threads enabled if needed.

(2) in a similar vein as https://github.com/MacPython/openblas-libs/pull/166

(3) included in https://github.com/MacPython/openblas-libs/pull/85 with the help of `-Wl,-gc-sections -Wl,-s` in the linking stage.

rgommers:
Re (1), the relevant issue is https://github.com/scipy/scipy/issues/15129. This isn't happening soon, it's blocked for at least two reasons (ILP64 vs. LP64, and Python packaging standards forbidding us from having an extra dependency in some wheels only).

(2) is being done.

(3) is always a good idea - if stripping isn't optimal yet, that's great to fix.

mattip:
The scipy-openblas-0.3.27.44.3 wheels, without #85, are [here](https://pypi.org/project/scipy-openblas64/0.3.27.44.3/#files). The win_amd64 one is 10.7MB.

The scipy-openblas-0.3.27.44.4 wheels, with #85, are [here](https://pypi.org/project/scipy-openblas64/0.3.27.44.4/#files). The win_amd64 one is 10.0 MB. Adding #85 and backing out a windows threading issue saved ~0.7MB. 

It seems #177 will shrink the wheel to 6.7MB. :tada:

rgommers:
That is excellent, thanks Matti. Also quite useful to have two tagged versions with the only change being the size change due to the dropped architectures - that's going to help in case we get some issue that may possibly be related.

mattip:
According to the documentation
```
-fno-ident
    Ignore the #ident directive.
```

Is there much use of that directive in gcc and/or OpenBLAS?

mattip:
[...]

PR #178, opened 2024-08-09
OpenBLAS v0.3.28

mattip:

- cherry-pick commit to build windows without quadmath, replaces #85
- update to OpenBLAS 0.3.28


mattip:
Downloading the windows wheels-x64-1 zipfile and opening it, the windows wheel is 10.0MB. I will try adding `-fno-ident` and see if that shrinks the wheel.

mattip:
... as mentioned in [this comment](https://github.com/MacPython/openblas-libs/issues/175#issuecomment-2277407895)

PR #181, opened 2024-08-11
refactor to acutally use DYNAMIC_LIST

mattip:

- don't redirect, add QUIET_MAKE to see if travis log file is still too big
- move some variables around

Trying this out locally, it seems I should see `-DDYN_PRESCOTT -DDYN_NEHALEM -DDYN_SANDYBRIDGE -DDYN_HASWELL -DDYN_SKYLAKEX` in the cc command line in the logs

### DIFF
--- a/tools/build_steps.sh
+++ b/tools/build_steps.sh
@@ -97,7 +97,7 @@ function patch_source {
     for f in $(ls ../patches); do
         echo applying patch $f
         git apply ../patches/$f
-    done 
+    done
 }
 
 function do_build_lib {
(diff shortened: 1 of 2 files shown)


tools/build_steps.sh
```diff
@@ -172,20 +172,18 @@ function do_build_lib {
...
+        make BUFFERSIZE=20 DYNAMIC_ARCH=1 QUIET_MAKE=1 \
             USE_OPENMP=0 NUM_THREADS=64 \
-            BINARY=$bitness $interface_flags $target_flags shared 2>&1 1>/dev/null
+            BINARY="$bitness" $interface_flags \
+            TARGET="$target"
     fi
-    echo done building, now testing
-    make BUFFERSIZE=20 DYNAMIC_ARCH=1 \
-        USE_OPENMP=0 NUM_THREADS=64 \
-        BINARY=$bitness $interface_flags $target_flags tests
```
mattip, line 188: I think this may have been the problem. I did it to break up the build and try to get some log files, since on TravisCI there is a hard limit on the size of the logs. But `make test` requires `make shared`, and the command here was without the `DYNAMIC_LIST`. So `make` rebuilt all the kernels :(.

mattip:
Comfirmed this time, the manylinux_x86_64 wheels have a 21MB shared object instead of a 35MB one.

PR #192, opened 2025-04-03
Add build script for Windows on ARM64

Harishmcw:
This PR adds `build_steps_win_arm64.bat`, a script that automates the process of building OpenBLAS and packaging it into a SciPy-compatible wheel on Windows ARM64. It compiles OpenBLAS using the ARM64 native command prompt, prepares the necessary files, and generates a wheel for seamless integration with SciPy.

### DIFF
--- a/tools/build_steps_win_arm64.bat
+++ b/tools/build_steps_win_arm64.bat
@@ -1,6 +1,6 @@
 :: Build script for scipy_openblas wheel on Windows on ARM64
 
-:: Usage: build_openblas.sh [build_bits]
+:: Usage: build_steps_win_arm64.bat [build_bits]
 :: e.g build_steps_win_arm64.bat 64
 
 :: BUILD_BITS  (default binary architecture, 32 or 64, unspec -> 64).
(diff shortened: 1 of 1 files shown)


mattip (commented):
Shouldn't the BUILD_BITS be used to define a prefix/suffix in the exported function names? Or am I missing where this is done?

Harishmcw:
> Shouldn't the BUILD_BITS be used to define a prefix/suffix in the exported function names? Or am I missing where this is done?

The suffix/prefix on exported symbols is determined based on the `if_bits` variable in the build_openblas.sh script, while the build_bits variable is used for configuring the OpenBLAS build.

If INTERFACE64 is set to 1, if_bits is assigned 64; otherwise, it defaults to 32. Since INTERFACE64 is not used here, if_bits remains 32, resulting in only the prefix scipy_ being added without any suffixes

mattip:
Ahh, I see. NumPy uses the 64-bit interfaces, so I guess more work will be needed to produce those wheels.

PR #193, opened 2025-04-10
Refactor Windows on ARM build script

matthew-brett:
Use the standard repo structure for build.

### DIFF
--- a/tools/build_steps_win_arm64.bat
+++ b/tools/build_steps_win_arm64.bat
@@ -7,6 +7,11 @@
 :: Expects these binaries on the PATH:
 :: clang-cl, flang-new, cmake, perl
 
+:: First commit containing WoA build fixes.
+:: Minimum OpenBLAS commit to build; we'll update to this if commit not
+:: present.
+set first_woa_buildable_commit="de2380e5a6149706a633322a16a0f66faa5591fc"
+
 @echo off
 setlocal enabledelayedexpansion
 
@@ -18,31 +23,24 @@ if "%1"=="" (
 echo Building for %BUILD_BIT%-bit configuration...
  
 :: Define destination directory
-move "..\local\scipy_openblas64" "..\local\scipy_openblas32"
-set "DEST_DIR=%CD%\..\local\scipy_openblas32"
-cd ..
+set ob_out_root=%CD%\local\scipy_openblas
+set ob_64="%ob_out_root%64"
+set ob_32="%ob_out_root%32"
+if exist %ob_64% copy %ob_64% %ob_32%
+set DEST_DIR=%ob_32%
+
+:: Clone OpenBLAS
+echo Cloning OpenBLAS repository with submodules...
+git submodule update --init --recursive OpenBLAS
+if errorlevel 1 exit /b 1
  
-:: Check if 'openblas' folder exists and is empty
-if exist "openblas" (
-    dir /b "openblas" | findstr . >nul
-    if errorlevel 1 (
-        echo OpenBLAS folder exists but is empty. Deleting and recloning...
-        rmdir /s /q "openblas"
-    )
+:: Enter OpenBLAS directory and checkout buildable commit
+cd OpenBLAS
+git merge-base --is-ancestor %first_woa_buildable_commit% HEAD 2>NUL
+if errorlevel 1 (
+    echo Updating to WoA buildable commit for OpenBLAS
+    git checkout %first_woa_buildable_commit%
 )
- 
-:: Clone OpenBLAS if not present
-if not exist "openblas" (
-    echo Cloning OpenBLAS repository with submodules...
-    git clone --recursive https://github.com/OpenMathLib/OpenBLAS.git OpenBLAS
-    if errorlevel 1 exit /b 1
-)
- 
-:: Enter OpenBLAS directory and checkout develop branch
-cd openblas
-git checkout develop
- 
-echo Checked out to the latest branch of OpenBLAS.
  
 :: Create build directory and navigate to it
 if not exist build mkdir build
@@ -76,7 +74,7 @@ powershell -Command "(Get-Content 'local\scipy_openblas32\__init__.py') -replace
 :: Prepare destination directory
 cd OpenBLAS/build
 echo Preparing destination directory at %DEST_DIR%...
-if not exist "%DEST_DIR%\lib\cmake\openblas" mkdir "%DEST_DIR%\lib\cmake\openblas"
+if not exist "%DEST_DIR%\lib\cmake\OpenBLAS" mkdir "%DEST_DIR%\lib\cmake\OpenBLAS"
 if not exist "%DEST_DIR%\include" mkdir "%DEST_DIR%\include"
  
 :: Move library files


Harishmcw:
Hi Mathew,

**Fixes in the Refactored Code:**

1. Fixing Directory Handling for Build Output: (ln No. 26)

```
pushd "%~dp0\.."
set "ob_out_root=%CD%\local\scipy_openblas"
set "ob_64=%ob_out_root%64"
set "ob_32=%ob_out_root%32"
if exist "%ob_64%" xcopy "%ob_64%" "%ob_32%" /I /Y
set "DEST_DIR=%ob_32%" 
```

- Adding pushd "%~dp0\.." makes the script more robust—it always sets the working directory to the root of the openblas-libs repo, no matter where the script is invoked from. 
- The original copy command was treating the folder as a file, leading to an error during execution. Switching to xcopy ensures proper copying of the folder and its contents.  

2. **Ensuring a Clean Build Environment: (ln. No 46)**  

```
if exist build (rmdir /S /Q build || exit /b 1)
mkdir build || exit /b 1 & cd build || exit /b 1 
```

- Prevents stale or conflicting build artifacts from affecting the current build. 
- Helps in achieving a clean and reproducible build every time.  

3. Adding a check before running Python -m build: (ln. No 114) 

`python -c "import build" 2>NUL || pip install build`

4. Adding the .h extension to the missing header file: (ln. No 100) 

`if exist lapacke_mangling.h copy /Y lapacke_mangling.h "%DEST_DIR%\include`

matthew-brett:
Thanks - could you check I have made those changes as you intended them?  By the way, I think you can make pull request to my branch with the changes directly.

Harishmcw:
Hi Mathew,

I was going through the bat script and noticed the environment variables are currently being set using inline quotes (e.g., set var="value"), which embeds the quotes inside the variable value. So the actual variable value is like: 
`"C:\some path\local\scipy_openblas64"`. This can lead to double quotes in unexpected places later in the script.

A safer and more robust approach would be:
```
set "ob_out_root=%CD%\local\scipy_openblas"
set "ob_64=%ob_out_root%64"
set "ob_32=%ob_out_root%32"
if exist "%ob_64%" xcopy "%ob_64%" "%ob_32%" /I /Y
set "DEST_DIR=%ob_32%" 
```
This version quotes the values correctly and handles paths with spaces or special characters safely.

Thanks!

matthew-brett:
Current code correct?

Harishmcw:
> Current code correct?

Yep, looks perfect now.👍

matthew-brett:
@Harishmcw - can you think of a way to avoid leaving behind a modified `pyproject.toml` file?

Harishmcw:
> @Harishmcw - can you think of a way to avoid leaving behind a modified `pyproject.toml` file?

Hi @matthew-brett ,

To avoid leaving a modified pyproject.toml file, we can create a backup of it.

Add this line after ln No.68:

`copy pyproject.toml pyproject.toml.bak`

This creates a backup of the original pyproject.toml file as pyproject.toml.bak before any modifications are made. This ensures that we have a copy of the original file.

Add this line after ln No.117:

`move /Y pyproject.toml.bak pyproject.toml`

This line restores the original pyproject.toml file from the backup (pyproject.toml.bak). The /Y option automatically confirms the overwrite, ensuring that the original file is restored and no modified version is left behind.

Let me know if you need any further clarification or adjustments!

Thanks!

matthew-brett:
How about the current code?

Harishmcw:
> How about the current code?

Yes, the current code works as expected. However, it does leave behind the pyproject_64_32.toml file after the process.

matthew-brett:
> > How about the current code?
> 
> Yes, the current code works as expected. However, it does leave behind the pyproject_64_32.toml file after the process.

I don't see that - I think the `move` commands take care of that deletion.

mattip:
It seems github has [opened the windows11-arm64 runners](https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/) for public repos like this. Want to add a workflow?

matthew-brett:
Recording progress here.

I've started a `.github/workflows/windows-arm.yml` runner config.

@vksukumarx worked out how to install LLVM directly via the installer (see the runner config).

The current problem is that the `cmake` installed on the WoA runner is configured for x86_64 - see https://github.com/MacPython/openblas-libs/actions/runs/14864637680/job/41738232174 for error:

> "C:/a/openblas-libs/openblas-libs/OpenBLAS/kernel/x86_64/KERNEL.ARMV8"
  cannot be read. 

This is a known problem that will apparently be fixed in the next version of the runner : 

https://github.com/actions/partner-runner-images/issues/84

I have tried using `winget` to install `cmake` via the usual CMake installer, but this fails because the installer fails on trying to uninstall the previous version of CMake - I'll get links to errors if anyone is interested. 

I have also tried to solve this by installing `cmake` via `msys2` on the image, over at https://github.com/matthew-brett/woa-setup - but this is failing due to permission errors, that I can reproduce on the laptop, but that I do not understand.   Here's the command I am using to install msys2 (after downloading the installer):

```
.\msys2_installer.exe install --root C:\msys2 --default-answer --confirm-command
```

but, no matter what directory I try, this always give me:

```
[100] Critical: Error writing Maintenance Tool:  "Cannot write installer configuration to C:/msys2/: Access error"
[100] WriteError : Error writing Maintenance Tool : Cannot write installer configuration to C:/msys2/: Access error
```

Someone else [ran into the same error on the ARM64 image](https://github.com/actions/runner-images/issues/11570).

There is an [open request to install msys2 on the runner](https://github.com/actions/partner-runner-images/issues/88) but it's not whether that will happen.

Any other suggestions for a working native CMake welcome.  Msys2 might also be a good way to get `clang` and `flang-new` (as they are built for msys2 and WoA).

To make installs easier, I have also asked the team to install `winget` on the runner image, to make installation easier:

https://github.com/actions/partner-runner-images/issues/95

although, as noted there, I have worked out how to do this a workflow, with considerable experimentation.

Suggestions?

rgommers:
`pip install cmake`? There's a `win-arm64` wheel up at https://pypi.org/project/cmake/#files

matthew-brett:
> `pip install cmake`? There's a `win-arm64` wheel up at https://pypi.org/project/cmake/#files

Thanks - good thought - with that unblocked, I worked through the rest, the build is now passing.

My remaining worries are:

* We don't appear to be building static and dynamic library link files (for `gfortran`, these would be `.a` and `.dll.a`.   We only have a `.LIB` file (and the `.dll`).   Am I missing something?
* I'm afraid I'm confused by the code in `tools/build_steps_win_arm64.bat` where we munge `pyproject.toml` and other files to replace `openblas64` with `openblas32`, regardless of the `BUILD_BITS`.  Shouldn't the naming reflect the `BUILD_BIT` variable?   For example, why do we end up with a `scipy_openblas32` wheel from a `BUILD_BIT=64` build?

Harishmcw:
> * I'm afraid I'm confused by the code in `tools/build_steps_win_arm64.bat` where we munge `pyproject.toml` and other files to replace `openblas64` with `openblas32`, regardless of the `BUILD_BITS`.  Shouldn't the naming reflect the `BUILD_BIT` variable?   For example, why do we end up with a `scipy_openblas32` wheel from a `BUILD_BIT=64` build?

Hi @matthew-brett,

To clarify:

- The `if_bits` variable, defined in ⁠[tools/build_openblas.sh](https://github.com/MacPython/openblas-libs/blob/29ffffbeb8b9609cf1dd22828f6dd6a927cd25c1/tools/build_openblas.sh#L29), determines whether the resulting wheel is named scipy_openblas32 or scipy_openblas64. On the other hand, the `build_bits` variable is used to set the target architecture through the -DBINARY flag in the CMake configuration when building OpenBLAS (i.e., whether to build a 32-bit or 64-bit library)

- Since SciPy currently uses the `scipy_openblas32` wheel (with 32-bit integer interfaces), the existing code is written to produce that. If there's a need to generate a `scipy_openblas64` wheel for NumPy (i.e., with `INTERFACE64=1`), then we'd need to modify the build logic to support that interface configuration explicitly.

- I'll be pushing the refactored bat script shortly that also supports setting the `INTERFACE64` flag, which will produce the `scipy_openblas64` wheel.

matthew-brett:
For my own understanding - is this correct?

The builds can be 32-bit or 64-bit builds - meaning - binaries for 32-bit for 64-bit architecture.   Call this "binary-bits".   In e.g. `tools/build_openblas.sh`, the variable `BUILD_BITS` specifies whether the binary-bits are 32 or 64.

The builds can have 64- or 32-bit API interfaces to BLAS / LAPACK.  Call this "API-bits".  `INTERFACE64=0` in the scripts gives API-bits of 32, `INTERFACE64=1` gives API-bits of 64.

The `pyproject.toml` files in the repository are designed for API-bits=64, and so the files have to be rewritten to build API-bits-32.

Is that right?

It doesn't make sense to have binary-bits = 32 and API-bits = 64.

Windows ARM effectively only exists in 64-bit binary interface form.

The current code builds binary-bits=64 and API-bits=32.

matthew-brett:
I renamed `BUILD_BIT` in `tools/build_steps_win_arm64.bat` to `BUILD_BITS` for clarity.

Harishmcw:
> The current WoA build script builds binary-bits=64 and API-bits=32.

Yes, that's correct — the current WoA build script builds with binary-bits = 64 and API-bits = 32 (i.e., INTERFACE64=0).

mattip:
Something seems off with the wheel zipped into [the build artifact](https://github.com/MacPython/openblas-libs/actions/runs/14913359802?pr=193/). I see a single wheel `scipy_openblas32-0.3.29.0.0-py3-none-any.whl`. A wheel is a zip file, so unzipping it I see
three directories: `scipy_openblas32`, `scipy_openblas32-0.3.29.0.0.dist-info`, and   `scipy_openblas64`. The last one is not supposed to be there. Some of the metadata in the `dist-info` directory mentions also `scipy_openblas64`, I guess if the directory is deleted/renamed before the wheel build it will not be mentioned.

I assume the wheel can be used to build scipy/numpy?

I am fine with merging this as-is, it would be nice to provide an win-arm64 wheel with 64-bit interfaces for NumPy, which would enable using larger arrays.

rgommers:
It also shouldn't be a `none-any` wheel, that's going to give problems probably; the `any` should be `win_arm64`.

Harishmcw:
Hi @mattip 
> Something seems off with the wheel zipped into [the build artifact](https://github.com/MacPython/openblas-libs/actions/runs/14913359802?pr=193/). I see a single wheel `scipy_openblas32-0.3.29.0.0-py3-none-any.whl`. A wheel is a zip file, so unzipping it I see three directories: `scipy_openblas32`, `scipy_openblas32-0.3.29.0.0.dist-info`, and `scipy_openblas64`. The last one is not supposed to be there. Some of the metadata in the `dist-info` directory mentions also `scipy_openblas64`, I guess if the directory is deleted/renamed before the wheel build it will not be mentioned.
> 

- Previously, while building the `scipy_openblas32` wheel, we were copying the contents of `scipy_openblas64` into a new directory named `scipy_openblas32`. However, since `scipy_openblas64` was still present in the local directory during the build process, running python -m build included everything in the local directory into the final wheel. This unintentionally added the `scipy_openblas64` directory and associated metadata to the wheel.
- We have now addressed this issue for interface 32-bit builds, where we temporarily move the `scipy_openblas64` folder out of the build context before packaging and restore it afterward. This prevents unrelated directories from being included in the generated wheel, resulting in a clean and correct build output.

>I assume the wheel can be used to build scipy/numpy?
>

- Yes, the code has been updated to support both 32-bit and 64-bit interface builds, and it ensures that the generated wheel is correctly tagged with `win_arm64`.
- I've attached the [Fixes.patch](https://github.com/user-attachments/files/20228260/Fixes.patch) file, which contains the updated code. This patch should be applied to the current code in this draft PR. **Note:** I was unable to attach the .bat file directly here, so I've included the patch file instead.

- During the 64-bit interface build, when we pass the `libnamesuffix` flag as 64_, the resulting library is named as `scipy_openblas64__64`. This happens because, in the CMake-based build of OpenBLAS, the [SUFFIX64_UNDERSCORE](https://github.com/OpenMathLib/OpenBLAS/blob/cf9e34c1f4e5702061e81f571c45c37d9c2c749f/cmake/system.cmake#L355-L358) flag is enabled by default when INTERFACE64 is set, which appends an additional _64 suffix. As a result, the library name ends up with a double suffix.

- To avoid this, we ignore the libnamesuffix flag during the build. The default suffix `_64` is then added by the OpenBLAS build system, giving us the library name `scipy_openblas_64`. While this doesn’t match our preferred naming convention `(scipy_openblas64_)`, it avoids the double suffix issue.

- We've raised an [issue](https://github.com/OpenMathLib/OpenBLAS/issues/5259) in the OpenBLAS repository to highlight this behavior, and we will raise a PR to adjust the `SUFFIX64_UNDERSCORE` flag so that the library is named as expected: `scipy_openblas64_`.

mattip:
I committed your change, let's see how it works

mattip:
@matthew-brett, @rgommers this looks got to me. Do you want to take a look?

.github/workflows/windows-arm.yml
```diff
@@ -0,0 +1,97 @@
...
+        path: dist/scipy_openblas*.whl
+
+    - uses: actions/upload-artifact@v4.3.0
+      with:
+        name: openblas-${{ env.PLAT }}-${{ env.INTERFACE64 }}
+        path: builds/openblas*.zip
+
+    - name: Install Anaconda client
+      run: |
+        Invoke-WebRequest https://static.rust-lang.org/rustup/dist/aarch64-pc-windows-msvc/rustup-init.exe -UseBasicParsing -Outfile rustup-init.exe
```
rgommers, line 84: It'd be useful to add a comment here about why this was necessary. Looks like it's only because of `rpds-py` (https://github.com/crate-py/rpds/issues/116)

rgommers (approved):
LGTM, just one minor comment. CI log looks as expected (modulo some log pollution from compiler warnings coming from reference LAPACK).

.github/workflows/windows-arm.yml
matthew-brett: That's right - noted.

tools/build_steps_win_arm64.bat
```diff
@@ -1,52 +1,83 @@
...
+:: build_bits  (default binary architecture, 32 or 64, unspec -> 64).
+:: if_bits (default interface size, 32 or 64, unspec -> 32)
+:: If INTERFACE64 environment variable is 1, then if_bits defaults to 64
 :: Expects these binaries on the PATH:
 :: clang-cl, flang-new, cmake, perl
 
+:: First commit containing WoA build fixes.
+:: Minimum OpenBLAS commit to build; we'll update to this if commit not
+:: present.
+set first_woa_buildable_commit="de2380e5a6149706a633322a16a0f66faa5591fc"
```
matthew-brett, line 15: Are we in a position to update the submodule OpenBLAS to some commit that contains the WoA build updates as above?   Otherwise - what should we do with the noted version if we have to do the update above?

matthew-brett:
In general, the replace / copy / move stuff for the 32 vs 64 bit seems like a hack - I mean in the Windows build in general, not just here.  Should we have a templated `pyproject.toml` instead, as in `pyproject.toml.tpl` and then fill it with the relevant values in a template rendering step, to give the definitive `pyproject.toml`?

tools/build_steps_win_arm64.bat
Harishmcw: Not at the moment — we'll update the submodule once the SUFFIX64_UNDERSCORE issue is fixed upstream. That way, we produce the correct libname for 64-bit interface builds as scipy_openblas64_.

matthew-brett: @Harishmcw - yes, we do not use a compatible OpenBLAS at the moment, but the question was whether we could update to one as part of this change.  Otherwise, I guess we could cherry-pick the build changes, so that the user would get the right version (e.g. v0.3.29) in terms of execution results.

Harishmcw: > @Harishmcw - yes, we do not use a compatible OpenBLAS at the moment, but the question was whether we could update to one as part of this change. Otherwise, I guess we could cherry-pick the build changes, so that the user would get the right version (e.g. v0.3.29) in terms of execution results.

The current OpenBLAS submodule commit we're using already includes all the Win ARM64-related changes, so no update is needed at this point.

matthew-brett: @Harishmcw - if you look at the code in `tools/build_openblas.sh`, you will see that it does `git checkout $OPENBLAS_COMMIT` where `OPENBLAS_COMMIT` comes from the top of the the `windows.yml` and `posix.yml` Github workflow files.   Currently this commit is `v0.3.29`.  This is so we can assure ourselves that we are building a specific version, that is constant across the platforms.   The WoA build recipe does not do this - but probably should.  The problem is the lack of WoA build patches in OpenBLAS v0.3.29 - hence my question above.

mattip: It is fine to change the `OPENBLAS_COMMIT` in all the workflow files to a newer commit. Then the version in the `pyproject.toml` should be adjusted to match, using `git describe --tags --abbrev=8 in the OpenBLAS directory. I thought I had a test to make sure the result of [get_openblas_config()](https://github.com/MacPython/openblas-libs/blob/main/local/scipy_openblas64/__init__.py#L116) will match the pyproject.toml version, but it seems not. :(

matthew-brett: @mattip - any preference for the newer (post WoA fixes) commit?   Current `develop` for example?

And then we can change the logic to error if the OpenBLAS commit doesn't contain the WoA fixes.

mattip: Any commit is fine, as long as the first fields of the version matches ``git describe --tags --abbrev=8`, then tag on a `0` for the wheel version that we can update if there is a problem with the wheel

matthew-brett: I've updated to the current HEAD of the OpenBLAS `develop` branch, but I wasn't sure where the `0` goes - can you advise?

pyproject.toml
```diff
@@ -9,7 +9,7 @@ build-backend = "setuptools.build_meta"
 [project]
 name = "scipy-openblas64"
 # v0.3.29
-version = "0.3.29.0.0"
+version = "0.3.29.265.0"
```
mattip, line 12: > I've updated to the current HEAD of the OpenBLAS develop branch, but I wasn't sure where the 0 (version) goes

Here :). I think this should be good to go now?

mattip:
Looks good. The wheels are named `scipy_openblas*-0.3.29.265.0-py3-none-*.whl`, which is correct.

matthew-brett:
Nice - thanks @mattip

khmyznikov:
@mattip what is the ETA for PyPi availability?

PR #207, opened 2025-06-28
use gcc from homebrew

mattip:
- [X] I updated the package version in pyproject.toml and made sure the first 3 numbers match `git describe --tags --abbrev=8` in OpenBLAS at the `OPENBLAS_COMMIT`. If I did not update `OPENBLAS_COMMIT`, I incremented the wheel build number (i.e. 0.3.29.0.0 to 0.3.29.0.1)


mattip:
I can't get this to work. It seems somehow specifying CC messes with the cross-compilation.

rgommers:
Can I ask what motivated the attempt to try making this switch? Using Clang seems healthier on macOS, since that's how everything gets compiled and it avoids pulling in C/C++ runtime libs.

rgommers:
I can't see so quickly if `zgetrf` uses a BLAS function that uses SME, but if so it could be similar to https://github.com/numpy/numpy/pull/29223.

I also noticed that the macos-arm64 builds on this repo still use `macos-13` x86-64 runners, which probably doesn't make too much sense anymore and will make it hard or impossible to see this in CI. Switching to `macos-latest` would be useful either way.

PR #210, opened 2025-07-02
update to macos-latest

mattip:
- [X] I updated the package version in pyproject.toml and made sure the first 3 numbers match `git describe --tags --abbrev=8` in OpenBLAS at the `OPENBLAS_COMMIT`. If I did not update `OPENBLAS_COMMIT`, I incremented the wheel build number (i.e. 0.3.29.0.0 to 0.3.29.0.1)

This moves the macos builds from a x86_64 host to an arm64 host, hopefully it will Just Work

### DIFF
--- a/.github/workflows/posix.yml
+++ b/.github/workflows/posix.yml
@@ -159,12 +159,13 @@ jobs:
         name: openblas-${{ matrix.os }}-${{ matrix.PLAT }}-${{ matrix.INTERFACE64 }}-${{ matrix.MB_ML_LIBC }}-${{ matrix.MB_ML_VER }}
         path: libs/openblas*.tar.gz
 
-    - uses: conda-incubator/setup-miniconda@v3.1.1
+    - uses: conda-incubator/setup-miniconda@v3.2.0
       with:
         channels: conda-forge
         channel-priority: true
         activate-environment: upload
         miniforge-version: latest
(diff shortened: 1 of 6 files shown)


mattip:
CI is passing. It took a bit to work around limitations in the gfortran_install and multibuild subrepos, maybe we should switch to the scipy strategy of using the conda-forge `compilers` package instead

rgommers:
Ah you switched all the builds to `macos-latest`; I had thought you'd leave the x86-64 builds on `macos-13` because those are x86-64 runners. I have no idea if and when GitHub will deprecate those.



> maybe we should switch to the scipy strategy of using the conda-forge `compilers` package instead

It's not a whole lot better (SDKROOT messing around still required), and actually it looks like for arm64 we're pulling in some other `gfortran`:

https://github.com/scipy/scipy/blob/9ae478ddaec6e1a71ca31624199e1ea171835eec/tools/wheels/cibw_before_build_macos.sh#L48

rgommers:
The more impactful maintenance improvement is probably to switch to `cibuildwheel`

rgommers:
(I'll leave the merging up to you, since I didn't follow along very closely)

mattip:
Thanks for the review
> The more impactful maintenance improvement is probably to switch to cibuildwheel

Good idea, I will open an issue

PR #220, opened 2025-10-23
cleanup clean_code, remove repo_dir, checkout OpenBLAS commit in ci-before-build.sh

mattip:

### DIFF
--- a/ci-before-build.sh
+++ b/ci-before-build.sh
@@ -47,8 +47,11 @@
 set -xe
 
 if [[ "$NIGHTLY" = "true" ]]; then
+    pushd OpenBLAS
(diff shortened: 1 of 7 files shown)


mattip:
since the version in pyproject.toml is updated on nightly builds in ci-before-build.sh, make sure the git checkout is correct. Along the way clean up more code.

ci-before-build.sh
```diff
@@ -47,8 +47,11 @@
 set -xe
 
 if [[ "$NIGHTLY" = "true" ]]; then
+    pushd OpenBLAS
+    git checkout develop
```
mattip: This was missing

Issue #211, opened 2025-07-20
Use cibuildwheel in the build


mattip:
We should look at cibuildwheel and possibly removing the complicated scripting subrepos. We actually use very little of multibuild, mainly to build in dockers for linux. Instead, cibuildwheel could do this for us. We also use gfortran compilers from gfortran_install on macos, but the subrepo carries an unused achives directory. We could vendor the single gfortran_utils.sh script file instead.

mattip:
> Could you tell me where `DOCKER_TEST_IMAGE` would be used

It is not used anymore. It can be removed.

> I'm quite concerned about whether my refactored wheels can maintain consistent behavior

The real test is if the wheels can be used by NumPy/SciPy. This requires building the wheels and then testing them in the downstream user's CI. The price of a bad version is quite low, since NumPy/SciPy pin to a specific version of the wheels. If we really wanted to test the wheels, we could add here 
- get an sdist of NumPy
- create a venv
- install the dependecies
- install the 32-bit and 64-bit wheels,
- for each variant
  - `spin build --with-scipy-openblas=[64 or 32]`
  - `spin test`

rgommers:
Most of those questions are for @mattip to answer, but a few comments:

> But in Windows CI, even if bash is specified, it may be switched to use pwsh instead. 

`shell: bash` should be reliable in GitHub Actions, but you need to specify it in every job step that needs it. 

> For example, on unsupported platforms, can we ensure that wheels built from sdist remain usable?

There isn't, nor should there be, an sdist. Uploading one would definitely not be reliable, and isn't all that useful since these wheels are mostly meant for vendoring by numpy/scipy rather than direct usage.

Issue #219, opened 2025-10-23
Lots of warnings when building macos x86_64, then it crashes when using the wheel


mattip:
After #217, which was merged the macos x86_64 wheels crash. I see many [warnings](https://github.com/MacPython/openblas-libs/actions/runs/18717594321/job/53380921379#step:5:11471) during the build like
```
ld: warning: reducing alignment of section __TEXT,__text \
    from 0x10000 to 0x1000 because it exceeds segment maximum alignment
```

When using the wheel in numpy/numpy#30049, there is a segfault. I can reproduce locally on a macbook M1 (arm64) by
```
arch -x86_64 bash
python3.12 -m venv /tmp/venv312x86_64
. /tmp/venv312x86_64/bin/activate
cd numpy
python -m pip install -r requirements/build_requirements.txt
python -m pip install scipy-openblas64-0.3.30.0.4
spin build --clean --with-scipy-openblas=64 -- -Dpkg_config_path=$PWD/.openblas
spin python -- -c "import numpy"
```

Then in the console app (launcher -> console) I see this log for Python:
```
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGBUS)
Exception Codes:       KERN_PROTECTION_FAILURE at 0x0000000112443740
Exception Codes:       0x0000000000000002, 0x0000000112443740

Termination Reason:    Namespace SIGNAL, Code 10 Bus error: 10
Terminating Process:   exc handler [92470]

VM Region Info: 0x112443740 is in 0x111e88000-0x114ab1000;  bytes after start: 6010688  bytes before end: 40294591
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      VM_ALLOCATE (reserved)      10f004000-111004000    [ 32.0M] rw-/rwx SM=NUL  reserved VM address space (unallocated)
      GAP OF 0xe84000 BYTES
--->  __TEXT                      111e88000-114ab1000    [ 44.2M] r-x/rwx SM=COW  /tmp/*/libscipy_openblas64_.dylib
      __DATA                      114ab1000-114ad8000    [  156K] rw-/rwx SM=COW  /tmp/*/libscipy_openblas64_.dylib

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libscipy_openblas64_.dylib    	       0x112443790 dcopy_k_NEHALEM + 80
[...]

PR #221, opened 2025-10-23
cibuildwheel for windows

ffgan:
see https://github.com/MacPython/openblas-libs/issues/211#issuecomment-3430874451
---

The CI results show that everything has passed. You can view the details at https://github.com/ffgan/openblas-libs/actions/runs/18747245828
I have merged the `win` and `win on arm` configurations into `posix.yml` and renamed that file to `wheel.yml`.
Regarding travis, since I cannot debug locally, I may add related commits in the PR to pass travis.

---

**other info**
Co-authored by: [nijincheng@iscas.ac.cn](mailto:nijincheng@iscas.ac.cn);

---

### DIFF
--- a/.github/workflows/posix.yml
+++ b/.github/workflows/posix.yml
@@ -214,6 +214,7 @@ jobs:
           cp for_test\test*.exe builds
 
     - name: prepare wheel
+      if: ${{ matrix.os == 'windows-latest' }}
       run: |
         ./tools/build_wheel_prepare_for_win.sh
 


mattip:
Thanks! Rebased on top of #220

mattip:
Sure, I disabled actions. Does the travis workflow work locally when run in a docker container with the appropriate `--platform` argument?

ffgan:
> Sure, I disabled actions. Does the travis workflow work locally when run in a docker container with the appropriate `--platform` argument?

It can run locally, but I tried it and encountered issues like illegal instruction errors. My local machine is fedora + x64. For s390x build, I running s390x containers with QEMU, illegal instruction problems are difficult to troubleshoot.

What's even more troublesome is the environment issue - it's relatively difficult to reproduce the travis CI environment locally. For example, travis uses Python 3.10 by default. I just tested on travis and found that cibuildwheel 3.x doesn't actually support Python 3.10 (Or simply put, I didn't carefully check the documentation. ). I need more information before I can provide cibuildwheel support for travis.

ffgan:
> Is there a cibuildwheel option to get the logs to be open? It is frustrating to have to continually scroll to the end to get the next 30 lines in a thousand line log.

Perhaps this link will be useful to you, https://cibuildwheel.pypa.io/en/stable/options/#build-verbosity

ffgan:
> Switching to macos-15-intel made the build fail like in #216. I guess this is better than producing a faulty wheel, as in issue #219. It seems there is a problem with these builds. I wonder how upstream is building and testing macos-x86_64 since the macos-13 images are EOL.

I agree with your idea - compile-time errors are better than runtime errors. Let's try our best to figure out how to make the CI pass here.

mattip:
We should get #223 to pass before merging this. Anyway we need to think about combining windows and posix yml files.

ffgan:
Some simple descriptions.

After thinking it over, I found that simply putting the Windows part into POSIX isn't really ideal. However, for now at least, everything is uniformly using cibuildwheel, so let's just leave it as is for the time being. In some existing places, the default bash must be switched to use `pwsh`. If it's not `pwsh`, there will be some strange issues - for example, `make install` will have problems. Or there's that very strange issue about which bash environment it actually is, or the space issue in the bash path `Program Files`. Or whether it should be \ or /. In short, debugging this on Windows can only be described as a nightmare. But for now, getting CI to pass is already the best result of my attempts.

Regarding the fact that most of the scripts in posix.yml are only used for Windows, I think there are two directions we can go. One is to move the bash scripts directly out and put them back in the regular script area, not bringing them into CI here. The second is to split up the CI, using a method similar to the `use` approach in CI to call/initialize the Windows environment, which would preserve CI support very well. Due to time constraints, I haven't tried this yet. Actually, there's an even more troublesome issue - the cache problem. If there's no cache, I strongly advise against attempting to refactor the Windows scripts - each attempt takes at least an hour, and even the most patient person will be worn down mentally.

ffgan:
cc @mattip

PR #223, opened 2025-10-25
update OpenBLAS commit, use macos-15.intel

mattip:

Towards #219.
- Use macos-15-intel. As found in #221 and #216, this causes test failures.
- Update to latest OpenBLAS head commit 585e6d06802a which is v0.3.30-317. This means we can remove the patch since it was incorporated upstream.
- Remove some leftover cruft from the move away from multibuild

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,5 +2,5 @@
 	path = OpenBLAS
 	url = https://github.com/xianyi/OpenBLAS.git
 [submodule "gfortran-install"]
-   path = gfortran-install
-   url = https://github.com/MacPython/gfortran-install.git
+	path = gfortran-install
+	url = https://github.com/MacPython/gfortran-install.git


ffgan:
> > Might be pragmatic to see if the pre-cibuildwheel-conversion code has the same issue when run on macos-15-intel
> 
> Yes it did. See #216

I think if possible, this issue may need assistance from the @ openblas maintainers to investigate. I've tried using the patch here as well as compiling with the commits included in that [PR](https://github.com/OpenMathLib/OpenBLAS/pull/5478), but I'm still getting the same error.

ffgan:
If needed, I can create a separate repository that uses GitHub Actions to reproduce this bug and simplify the current workflow, so that the OpenBLAS team can investigate it more easily.

ffgan:
> I think we should reverse course in windows and macos. We should not use cibuildwheel there, the containerization in cibuildwheel only obfuscates the scripts. We should set up a build environment, probably the easiest is with conda compilers, and then call build-openblas.sh. This means we will need to bump the x86_64 MACOSX_DEPLOYMENT_TARGET from 10.9 to 10.13 (since that is what is supported in the conda compilers), but I think that is the least bad option.
> 
> I think it would simplify upstream compatibility if we use CMake in build-openblas.sh, but that is a separate issue.

Perhaps we could open these two issues. If possible, could you please assign the CMake one to me? I'll try to add support for it in a few days.

mattip:
- arm64 manylinux is failing tests with `SBGEMM FAILURES: 28800`, arm64 musllinux is passing. Strange. This is on OpenBLAS commit 585e6d06802a0 which is the current develop HEAD
- macos-x86_64 is still failing with `TEST 1/125 min:smin_negative make[1]: *** [run_test] Bus error: 10`, using xcode 16 and gfortran `GNU Fortran (GCC) 11.3.0`.
- macos-arm64 fails to build. There are problems with fortran linking. This is something new since commit 7cbbb0f. 
 
This whole PR is in worse shape than #220, which was passing CI. Maybe I should start again.

mattip:
Closing in favor of #225

Issue #170, opened 2024-07-29
Shrink aarch64 wheels


mattip:
I wonder if the problem with aarch64 builds on travisCI is that we are running out of memory and the build process is killed (on manylinux/glibc). Travis has [a 3GB limit](https://docs.travis-ci.com/user/common-build-problems/#my-build-script-is-killed-without-any-error). Similar to issue #144 and the PR #166, we should benchmark aarch64 on a high-end aarch64 machine.

@ev-br is this something you could do? Is the AWS m7g instance (with a graviton3 processor) advanced enough to use the `THUNDERX3T110` kernels or is that targeting some other processor?

Mousius:
The `THUNDERX3T110` target uses AdvSIMD only, whereas the `NEOVERSEV1` target on the AWS M7g can use SVE. Mostly the SVE targets remap back to `NEOVERSEV1` at the moment, so removing that would be pretty bad for performance.

Mousius:
I remapped any common targets back together in https://github.com/OpenMathLib/OpenBLAS/pull/4389, unsure how to tell which targets are less used and be removed 🤔 

Also ref: https://github.com/OpenMathLib/OpenBLAS/blob/develop/Makefile.system#L686-L700

mattip:
Thanks. Is `NEOVERSEV1` active [when using GCC](https://github.com/OpenMathLib/OpenBLAS/blob/6d071f1a1c13e71b67171179dd70f1d3728229fe/Makefile.system#L405C22-L405C25) (like in the build here)?

ev-br:
BLAS-benchmarks runs on a c7g.large instance (https://aws.amazon.com/ec2/instance-types/c7g/) via https://github.com/OpenMathLib/BLAS-Benchmarks/blob/main/.cirun.yml
Would this be enough?

Also, does @czgdp1807 benchmarking machinery handle aarch architectures?

Mousius:
> Thanks. Is `NEOVERSEV1` active [when using GCC](https://github.com/OpenMathLib/OpenBLAS/blob/6d071f1a1c13e71b67171179dd70f1d3728229fe/Makefile.system#L405C22-L405C25) (like in the build here)?

In manylinux2014 with GCC 10.2 you should get the SVE targets.

For certain toolchains, such as the `MACOSX_DEPLOYMENT_TARGET`, there isn't full support and it's disabled.

mattip:
[...]

PR #227, opened 2025-10-30
update OpenBLAS version to v0.3.30-322-gef6f9762

mattip:

Towards numpy/numpy#5520, scipy/scipy#23686. The last commit in OpenBLAS should fix them.


mattip:
The manylinux2014 image is using gcc-10 (and gfortran-10). The musllinux_1_2 image uses gcc/gfortran 14. gcc-10 is really out of date. FWIW, NumPy has dropped manylinux2014 in favor of manylinux_2_28 (manylinux2014 is equivalent to manylinux_2_17). Scipy still is releasing manylinux2014 wheels. We could
- use manylinux_2_28 for aarch64. This would then preclude using the updated wheel on scipy until they change their policy
- try to update gcc on the aarch64 manylinux2014 image. This will probably fail, according to [this comment](https://github.com/pypa/manylinux/issues/1266#issuecomment-1025197334) in the manylinux repo.
- Try to figure out which kernel is failing and disable it. I see on the [NumPy workflow](https://github.com/numpy/numpy/actions/runs/19000000408/job/54265358022#step:7:2241) that the runner machines support all the possible SIMD extensions `"NEON", "NEON_FP16", "NEON_VFPV4", "ASIMD", "ASIMDHP", "ASIMDDP", "ASIMDFHM", "SVE"`. I tried limiting the aarch64 kernels. We had started something like this in order to limit the package size in #170 but decided the ROI was not worth the effort.

PR #230, opened 2025-11-03
use qemu on github CI for ppc64le, s390x

mattip:
- [ ] I updated the package version in pyproject.toml and made sure the first 3 numbers match `git describe --tags --abbrev=8` in OpenBLAS at the `OPENBLAS_COMMIT`. If I did not update `OPENBLAS_COMMIT`, I incremented the wheel build number (i.e. 0.3.29.0.0 to 0.3.29.0.1)

Now that we have moved to cibuildwheel on linux, it should be easy to add ppc64le and s390x builds to the matrix. Let's see how long it takes to run. Travis CI is unstable, it would be nice to move off it.

### DIFF
--- a/.github/workflows/posix.yml
+++ b/.github/workflows/posix.yml
@@ -74,7 +74,7 @@ jobs:
       with:
         xcode-version: '15.4'
 
-	- name: Set up QEMU
+    - name: Set up QEMU
       if: runner.os == 'Linux' && runner.arch == 'X64'
       uses: docker/setup-qemu-action@v3
       with:


mayeut:
I wanted to check if something could be cross-build with clang but seeing QEMU was failing here, wanted to check that first.
- S390X: the samin/samax/damin/damax kernels are using the `lper`/ `lpder`instructions which are not supported by QEMU (SIGILL triggered). Once those specific tests are disabled, everything else passes (when tests don't hang but it should be fix by #227 ? or another OpenBLAS upgrade).
They can be disabled like the ones for loongarch64:
```diff
diff --git a/tools/build_steps.sh b/tools/build_steps.sh
index ee91642..c11ba72 100644
--- a/tools/build_steps.sh
+++ b/tools/build_steps.sh
@@ -220,9 +220,17 @@ EOF
         echo -n > utest/test_dsdot.c
         echo "Due to the qemu versions 7.2 causing utest cases to fail,"
         echo "the utest dsdot:dsdot_n_1 have been temporarily disabled."
+    elif [ "$plat" == "s390x" ]; then
+        sed -i 's/CTEST(samin, positive_step_1_N_70){/CTEST_SKIP(samin, positive_step_1_N_70){/g' ./utest/test_extensions/test_samin.c
+        sed -i 's/CTEST(samin, negative_step_1_N_70){/CTEST_SKIP(samin, negative_step_1_N_70){/g' ./utest/test_extensions/test_samin.c
+        sed -i 's/CTEST(damin, positive_step_1_N_70){/CTEST_SKIP(damin, positive_step_1_N_70){/g' ./utest/test_extensions/test_damin.c
+        sed -i 's/CTEST(damin, negative_step_1_N_70){/CTEST_SKIP(damin, negative_step_1_N_70){/g' ./utest/test_extensions/test_damin.c
+        echo "the utest samin/damin have been temporarily disabled."
+        echo "QEMU does not support the 'lper' /'lpdr' instructions used"
     fi
     if [ -n "$dynamic_list" ]; then
         CFLAGS="$CFLAGS -fvisibility=protected -Wno-uninitialized" \
```
- PPC64LE: OpenBLAS does not support customising the DYNAMIC_LIST which ends up including an x86_64 kernel...

The builds that did not hang on tests took around 3 hours: https://github.com/mayeut/openblas-libs/actions/runs/19394958005/job/55493703960

tools/build_steps.sh
```diff
@@ -181,11 +181,14 @@ EOF
...
             ;;
         *-s390x)
+            # The TargetList.txt has only ZARCH_GENERIC, Z13, Z14. Not worth
+            # messing with dynamic lists and targets.
             local bitness=64
             ;;
         *-ppc64le)
             local bitness=64
             local target="POWER8"
+            local dynamic_list="POWER8 POWER10"
```
mayeut:
```suggestion
```

ppc64le does not support dynamic list

tools/build_steps.sh
```diff
@@ -181,11 +181,14 @@ EOF
             export DYLD_LIBRARY_PATH=/usr/local/lib:$DYLD_LIBRARY_PATH
             ;;
         *-s390x)
+            # The TargetList.txt has only ZARCH_GENERIC, Z13, Z14. Not worth
+            # messing with dynamic lists and targets.
             local bitness=64
```
mayeut:
```suggestion
            # messing with dynamic lists.
            local bitness=64
            local target="ZARCH_GENERIC"
```
see https://github.com/OpenMathLib/OpenBLAS/blob/f6df9bebbb4259aa61ab5634c0f1269fb152cc0e/README.md?plain=1#L271-L273

tools/build_steps.sh
mattip: The [README](https://github.com/OpenMathLib/OpenBLAS/blob/f6df9bebbb4259aa61ab5634c0f1269fb152cc0e/README.md?plain=1#L263) seems to imply that DYNAMIC_LIST is supported
> For **POWER**, the list encompasses POWER6, POWER8 and POWER9. POWER10 is additionally available if a sufficiently recent compiler is used for the build.

mattip: Strange. Is there an open OpenBLAS issue about the missing `DYNAMIC_LIST` override of `DYNAMIC_CORE`?

Issue #233, opened 2025-11-20
Calling zladiv via ctypes segfaults on macos with v0.3.30.0.3 and greater


mattip:
Distilled from the failed test in scipy/scipy#23986 and scipy/scipy#23844

This code passes on linux, passes on macOS-arm64 with v0.3.30.0.2 and before, and segfaults (!) on macos-arm64 with v0.3.30.0.3 and greater. It uses ctypes to call zladiv with a complex double interface class, taken from [this stackoverflow answer](https://stackoverflow.com/questions/13373291/complex-number-in-ctypes).

Note the scipy code using cython wrappers does not crash, but does return a wrong answer.

```
import math
import ctypes as ct
import scipy_openblas32

# Prepare some ctype equivalents
class c_double_complex(ct.Structure): 
    """complex is a c structure
    https://docs.python.org/3/library/ctypes.html#module-ctypes suggests
    to use ctypes.Structure to pass structures (and, therefore, complex)
    """
    _fields_ = [("real", ct.c_double),("imag", ct.c_double)]
    @property
    def value(self):
        return self.real+1j*self.imag # fields declared above

c_double_complex_p = ct.POINTER(c_double_complex) # pointer to our complex

b_double_complex = lambda c: ct.byref(c_double_complex(c.real, c.imag))


blas = ct.CDLL(scipy_openblas32.dll._name)
blas.scipy_zladiv_.argtypes = [c_double_complex_p, c_double_complex_p]
blas.scipy_zladiv_.restype = c_double_complex

def test_complex_ladiv():
    cx = .5 + 1.j
    cy = .875 + 2.j
    ret = blas.scipy_zladiv_(b_double_complex(cy), b_double_complex(cx))
    assert math.isclose(ret.real, 1.95)
    assert math.isclose(ret.imag, 0.1)

test_complex_ladiv()
```

Something must have changed with the compilation flags?

Edit: ~the function signature here is wrong. When I figure out how to get ctypes to return a pointer to a struct as one of the arguments I will update the reproducer.~ I am pretty sure the signature is correct, it is mangled by one of the wrappers to return a value?

mattip:
[Here](https://github.com/OpenMathLib/OpenBLAS/blob/f6df9bebbb4259aa61ab5634c0f1269fb152cc0e/lapack-netlib/SRC/zladiv.c#L578) is the function definition. It is exported without an explicit declaration in a header file (as far as I can tell)
```C
[...]

PR #236, opened 2025-11-20
use gfortran 11

mattip:

Towards #233
Try using gfortran 11


mattip:
Yup, using gfortran-11 from https://github.com/isuruf/gcc/releases/download/gcc-11.3.0-2 solves #233 on macos-arm64

PR #237, opened 2025-11-20
always use gfortran from conda on macos, update to gfortran-15

mattip:

Port #236 to always use gfortran from conda, and also update to the newer packages (thanks @isuruf).


mattip:
Closing in favor of #240

PR #238, opened 2025-11-22
Use clang instead of gcc to build on Linux

mayeut:

Builds on top of #230 

The use of clang instead of gcc allows:
- to get a very recent compiler that supports recent SIMD extensions without having to wait for a new gcc-toolset or update the manylinux base image.
- to get faster builds when using QEMU

The clang install script might end-up included in manylinux images (see https://github.com/pypa/manylinux/discussions/1871) and has been copied directly from https://github.com/scikit-build/ninja-python-distributions/blob/master/scripts/install-static-clang.sh for now.

### DIFF
--- a/.github/workflows/posix.yml
+++ b/.github/workflows/posix.yml
@@ -106,6 +106,10 @@ jobs:
         CIBW_MUSLLINUX_X86_64_IMAGE: ${{ matrix.MB_ML_LIBC }}${{matrix.MB_ML_VER}}
         CIBW_MANYLINUX_AARCH64_IMAGE: ${{ matrix.MB_ML_LIBC }}${{matrix.MB_ML_VER}}
         CIBW_MUSLLINUX_AARCH64_IMAGE: ${{ matrix.MB_ML_LIBC }}${{matrix.MB_ML_VER}}
+        CIBW_MANYLINUX_PPC64LE_IMAGE: ${{ matrix.MB_ML_LIBC }}${{matrix.MB_ML_VER}}
+        CIBW_MUSLLINUX_PPC64LE_IMAGE: ${{ matrix.MB_ML_LIBC }}${{matrix.MB_ML_VER}}
+        CIBW_MANYLINUX_S390X_IMAGE: ${{ matrix.MB_ML_LIBC }}${{matrix.MB_ML_VER}}
+        CIBW_MUSLLINUX_S390X_IMAGE: ${{ matrix.MB_ML_LIBC }}${{matrix.MB_ML_VER}}
 
     - name: Upload wheels to artifacts
       uses: actions/upload-artifact@v4.3.0


mattip:
> I though the fork tests hang would disappear

One of the ppc64le runs succeeds, the other fails. The failed run prints
```
2025-11-23T08:09:19.4979912Z TEST 122/127 zgemv:2_0_nan_1_inf_1_incy_2 [OK]
2025-11-23T08:09:19.5031453Z TEST 123/127 potrf:bug_695 [OK]
2025-11-23T08:09:19.5115065Z TEST 124/127 potrf:smoketest_trivial [OK]
2025-11-23T08:09:19.7959236Z TEST 125/127 kernel_regress:skx_avx [OK]
2025-11-23T08:35:46.7868483Z ##[error]The action has timed out.
```

The successful run prints
```
2025-11-23T07:58:44.4121838Z TEST 122/127 zgemv:2_0_nan_1_inf_1_incy_2 [OK]
2025-11-23T07:58:44.4172979Z TEST 123/127 potrf:bug_695 [OK]
2025-11-23T07:58:44.4255760Z TEST 124/127 potrf:smoketest_trivial [OK]
2025-11-23T07:58:44.7111789Z TEST 125/127 kernel_regress:skx_avx [OK]
2025-11-23T07:59:10.0551932Z TEST 126/127 fork:safety [OK]
2025-11-23T07:59:10.0740217Z TEST 127/127 fork:safety_after_fork_in_parent [OK]
```

which suggests the problem is in `fork:safety`. 

The test itself is the one from [the scipy issue](https://github.com/scipy/scipy/issues/23686#issuecomment-3381958611) which is also the test in #229. I will try to debug it in a qemu docker container.

mattip:
Another problem: It seems this compiled shared object from the `wheels-macos-latest-arm64-1-macosx-` artifact suffers from the same segfault from issue #233 when testing the `zladiv` interface. Did something change in the way gfortran exports functions?

mayeut:
> It seems this compiled shared object from the wheels-macos-latest-arm64-1-macosx- artifact suffers from the same segfault from issue https://github.com/MacPython/openblas-libs/issues/233 when testing the zladiv interface. Did something change in the way gfortran exports functions?

This PR does not touch the macOS build except for the OpenBLAS update which only has a limited diff compared to what's in main, the only thing related to fortran is https://github.com/OpenMathLib/OpenBLAS/pull/5540 which seems right. Does main passes (or the current nightly build which uses the latest develop) ?

As a side note, this PR still uses gfortran on Linux.

mattip:
> I will try to debug it in a qemu docker container.

It is a little convoluted to reproduce the cibuildwheel build since it uses build isolation, so maybe I am not 1:1 accurate but:

When I run the make command without `QUIET_MAKE=1`, I see it is using `cc` as the C compiler. Even when setting `PATH=/opt/clang/bin:$PATH`, `cc` is `/opt/rh/gcc-toolset-14/root/usr/bin/cc`.

mayeut:
> Maybe we could patch the Makefile to print out the compiler locations and versions just to be sure we are using the right ones

It's done at the end - once the build succeeds - that's how I found out gfortran was not found on macOS arm64, we might want to ask for this to also be printed early on.

mattip:
It might be a qemu bug. I don't see anything relevant though in the [qemu issue tracker](https://gitlab.com/search?search=fork&nav_source=navbar&project_id=11167699&group_id=3038080&scope=issues&state=opened). Can you try running on a x86_64 host even though it is slower?

mayeut:
> Can you try running on a x86_64 host even though it is slower?

Already done in CI, the same happened: https://github.com/MacPython/openblas-libs/actions/runs/19600277946/job/56130767510 running on x86_64 (all 4 QEMU builds running on x86_64 deadlocked)

mattip:
The whole move to cibuildwheel has made this less stable, harder to understand, and harder to debug, starting with the inconvenience of looking at compacted build logs. Maybe we should go back to multibuild.

mattip:
I don't have qemu-user installed on the machine. I used `docker --platform`. 
```
$ docker --version
Docker version 23.0.0, build e92dd87
```

I do have 
```
$ qemu-ppc64le-static --version
qemu-ppc64le version 8.2.2 (Debian 1:8.2.2+ds-0ubuntu1.10)
```

mattip:
> However the fix I did not understand for https://github.com/MacPython/openblas-libs/pull/229 / https://github.com/OpenMathLib/OpenBLAS/issues/5520 does not seem to fix the issue reported upstream

What are you trying?

mattip:
Ahh, right, the patch in #229 was not a revert of the incorrect OpenMathLib/OpenBLAS#5170. Good catch.

pyproject.toml
```diff
@@ -53,4 +53,7 @@ environment-pass = [
   "BUILD_DIR",
   "PLAT",
   "OS-NAME",
+  "RUNNER_ARCH",
 ]
+[tool.cibuildwheel.linux]
+environment = { CC = "/opt/clang/bin/clang", CXX = "/opt/clang/bin/clang++", LDFLAGS = "-fuse-ld=lld" }
```
mattip: Do you need `FC` for gfortran (and maybe some FFLAGS)?

mattip:
Cool, CI is passing.

mattip:
Thanks @mayeut

mattip:
I wonder how hard it would be to cache the compiler tarballs and manylinux images

PR #239, opened 2025-11-24
use gfortran-15 via FC environment variable on macos

mattip:

The `alias gfortran=gfortran-15` did not work, as seen in the line `OpenBLAS: Detecting fortran compiler failed. Can only compile BLAS and f2c-converted LAPACK` hidden deep in the logs. I think this is because the alias does not pass command line arguments.


mattip:
This is duplicating much of #237, closing in favor of iterating on that

PR #267, opened 2026-03-19
move to manylinux_2_28 and use the distro's gcc14

mattip:
- [x] I updated the package version in pyproject.toml and made sure the first 3 numbers match  `./openblas_commit.txt`. If I did not update `./openblas_commit.txt`, I incremented the wheel build number (i.e. 0.3.29.0.0 to 0.3.29.0.1)

Note: update `./openblas_commit.txt` with `cd OpenBLAS; git describe --tags --abbrev=8 > ../openblas_commit.txt`

Move to manylinux_2_28 and use the gcc compiler there. This solves numpy/numpy#30816, as tested in numpy/numpy#30818 on aarch64. I hope none of the other platforms need clang.


mattip:
Strange. The build under qemu with the same compiler but using manylinux_2_28 instead of manylinux2014 times out after an hour, it used to finish in 26 minutes.

mattip:
Let's start again and only change the manylinux image.

PR #277, opened 2026-07-17
update to v0.3.34

mattip:
- [x] I updated the package version in pyproject.toml and made sure the first 3 numbers match  `./openblas_commit.txt`. If I did not update `./openblas_commit.txt`, I incremented the wheel build number (i.e. 0.3.29.0.0 to 0.3.29.0.1)

Note: update `./openblas_commit.txt` with `cd OpenBLAS; git describe --tags --abbrev=8 > ../openblas_commit.txt`


pyproject.toml
```diff
@@ -8,8 +8,8 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "scipy-openblas64"
-# v0.3.33-201-ge70a7a73
-version = "0.3.33.201.0"
+# v0.3.34.0
+version = "0.3.34.0"
```
mayeut:
```suggestion
version = "0.3.34.0.0"
```

Should the version always be 5 components ? The PR template suggest so.
i.e. the minor version of openblas followed by post commits (in this case 0 as it's using a tag) then this repo build version ?

mattip: Thanks, yeah, I messed this up

